PR #67, opened 2014-12-19
Feature: add 'per' request parameter to indicate page size

kookster:


app/controllers/api/accounts_controller.rb
```diff
@@ -6,12 +6,16 @@ class Api::AccountsController < Api::BaseController
 
   def resources
     @accounts ||= if params[:user_id]
-      User.find(params[:user_id]).accounts.order(created_at: :desc).page(params[:page])
+      User.find(params[:user_id]).accounts
     else
-      Account.active.member.order(created_at: :desc).page(params[:page])
+      Account.active.member
```
chrisrhoden: This looks like it might cause us some problems, no?

kookster: it is still doing the paging, it just does it now by default in the base class, with the specific ordering set by the method below `with_sorting`

kookster: the new `decorate_query` method calls `with_paging` by default, so the resource method doesn't have to.

app/controllers/api/picks_controller.rb
```diff
@@ -5,7 +5,7 @@ class Api::PicksController < Api::BaseController
   represent_with Api::PickRepresenter
 
   def resources
-    @picks = Pick.named_playlists.page(params[:page])
+    @picks = Pick.named_playlists
```
chrisrhoden: I don't think we want every pick ever to come through the API, am I misunderstanding this?

app/controllers/api/accounts_controller.rb
chrisrhoden: Ah, looks like you add it automatically later, got it.

app/controllers/api/picks_controller.rb
kookster: yes, base class handling this now, still paged via `with_paging`.

chrisrhoden: I think I see it later, disregard

app/controllers/concerns/hal_actions.rb
```diff
@@ -94,8 +93,23 @@ def resources_name
...
-    keys ||= []
+  def decorate_query(res)
+    res = with_sorting(res)
+    res = with_paging(res)
+    res = with_params(res)
+    res
+  end
+
+  def with_sorting(arel)
+    arel.order(id: :desc)
```
chrisrhoden: It's not clear to me that we want to sort by id by default, is there another reason I'm too dense to see why we're breaking this method out to be overridden at all, rather than just returning things sorted from the `resources` methods? If so, does it seem reasonable to have it just return its argument by default?

app/controllers/concerns/hal_actions.rb
```diff
@@ -59,11 +59,10 @@ def zoom_param
     }.call
   end
 
-
   def index_cache_path
     index_resources = self.try(:resources_base) || resources
     PagedCollection.new(
-      with_params(self.class.resources_params, index_resources),
+      decorate_query(index_resources),
```
chrisrhoden: Trying to get a handle on this, is [this bit](https://github.com/PRX/cms.prx.org/blob/71164ab63b3a78f93c9a26eb311cfaaaa15e7913/app/models/paged_collection.rb#L27) the thing that protects us from generating the same cache key for multiple collections containing the same most recently updated thing but different constituents? I suspect this becomes a greater concern when we start allowing the `per` parameter, so it is probably something to keep an eye on.

app/controllers/concerns/hal_actions.rb
```diff
@@ -59,11 +59,10 @@ def zoom_param
...
       request,
       item_class: self.class.resource_class,
       item_decorator: self.class.resource_representer
```
chrisrhoden: definitely not something to fix in this PR, but just noticing we can likely optimize away lots of object creation on a cache miss here by initializing only one PagedCollection (i.e. if `resources_base` is unavailable, short circuit to a memoized call to `index_collection`). Wanted to get it written down somewhere

chrisrhoden:
Looks good, only one real question and a note for something TODO in the future. I'm wondering if it's necessary to make sure we have good coverage on this behavior. My gut tells me it's probably pretty well tested against completely by the controller tests, but if you're down to write some units they'd be welcome.

app/controllers/concerns/hal_actions.rb
kookster: it is broken out so sorting can be specified without overriding `resources`.

as to the default, I hear ya, it was already defaulting to this, so this is not a change just moving where it happens to a better place:
``` ruby
# previously it did this
res = self.class.resource_class.order(id: :desc).page(params[:page])
```
https://github.com/PRX/cms.prx.org/pull/67/files#diff-05e2417b36696821063bd3631c843187L88

I was questioning it when I added it, but still lean to having it add something.

I would not be against changing it to do nothing by default (return the argument), but by id: :desc seemed like a more common case and so a more logical default.

app/controllers/concerns/hal_actions.rb
kookster: Yes. That is the part that creates a unique cache key based on the items in the collection, though I see I need to add a position integer to it so we handle order changes, but otherwise looks right.

kookster: http://jira.prx.org/browse/PRXNXT-928

kookster:
I'll add some more tests.
The other issues raised in the comments have now become jira tickets.

chrisrhoden:
@kookster nice

kookster:
@chrisrhoden I added a bunch of tests for hal_actions, should be good to go now.

PR #68, opened 2014-12-19
Feature: Persist audio files to stories

kookster:
Add in ability to create an `AudioFile` from an s3 upload url.
Use and track asynch tasks (via Fixer) to copy this from upload location and validate.
Handle callbacks for task completions.


kookster:
this is outdated - the new fixer client works very differently.

PR #69, opened 2015-01-05
Deploy to cms.prx.org

farski:


chrisrhoden:
@farski looks like cms.prx.org is not yet propagated, but you said yesterday you asked for the DNS records – I don't see a ticket in the request portal but it's likely that I don't get to see tickets filed by you. Are we still waiting for them to take care of this?

farski:
Still waiting; they may not have seen the request yesterday as I added it to the id. ticket, so I just opened a new ticket.

chrisrhoden:
@farski rgr that. This one (unlike auth/id) requires some server-side noodling to coincide with it so we should coordinate when we want to make an actual switch.

farski:
This DNS change seems to be generally available now

chrisrhoden:
@farski cms.prx.org is now live with hal.prx.org as an alias. If there is a timeline for us to shut down hal.prx.org or begin redirecting all traffic (may break some things, as net::http does not follow redirects by default) I'll leave that to you to keep in motion. Merging this for now.

farski:
Have you done the server-side things that needed to be done?

farski:
I switched over the categorizer to hit cms.prx.org. Was anything else other than the frontend hitting hal.prx.org?

chrisrhoden:
The PMP importer, I'd check in with @kookster to confirm

PR #71, opened 2015-01-22
Credit should not return caption

farski:


app/models/image.rb
```diff
@@ -14,7 +14,7 @@ def caption
   end
 
   def credit
-    self.try(:read_attribute, :caption)
+    self.try(:read_attribute, :credit)
```
chrisrhoden: Redundant `try` detected.

Redundant method detected.

farski: This is maybe a change for a different PR?

farski:
@chrisrhoden @kookster Any thoughts on this? It's holding up https://github.com/PRX/www.prx.org/pull/165 so maybe we can ticket the refactor but get the change through?

app/models/image.rb
chrisrhoden: Yep, disregard.

chrisrhoden:
@farski I have nothing to add, not sure if there's a reason for this to behave the old way that maybe @kookster can elucidate.

@kookster was there a reason that `Image#credit` was set to return `#caption`, or was that an oversight?

kookster:
@chrisrhoden I am sure that is an oversight - at the time nothing was using those attributes in the API, and I didn't notice the bug.

I am not sure I agree the minimal refactor @chrisrhoden proposed should be put in another ticket and PR, but I will take up that discussion on slack.

kookster:
after much discussion, I'm not sure we have resolution on what to do about minor stuff, but I think it should be fixed, and can confirm @chrisrhoden was right that the `self.try` were there b/c some of the image tables have caption and credit, and some don't.

Here's what I think - we need to fix that problem, these many image tables should be consistent, and then these methods won't be necessary at all. That is without a doubt another ticket, and I'll log that, and we can move on.

PR #74, opened 2015-02-11
Use duration rather than length for musical works

farski:
Having a `length` property does not seem to play well with the front end

### DIFF
--- a/app/representers/api/musical_work_representer.rb
+++ b/app/representers/api/musical_work_representer.rb
@@ -9,7 +9,7 @@ class Api::MusicalWorkRepresenter < Api::BaseRepresenter
   property :label
   property :album
   property :year
-  property :excerpt_length, as: :length
+  property :excerpt_length, as: :duration
 
   def self_url(musical_work)
     polymorphic_path([:api, musical_work.story, musical_work])
--- a/test/representers/api/musical_work_representer_test.rb
+++ b/test/representers/api/musical_work_representer_test.rb
@@ -17,7 +17,6 @@ describe Api::MusicalWorkRepresenter do
     json['id'].must_equal musical_work.id
   end
 
-
   it 'will have a nested self link' do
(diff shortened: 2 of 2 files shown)


app/representers/api/musical_work_representer.rb
```diff
@@ -9,7 +9,7 @@ class Api::MusicalWorkRepresenter < Api::BaseRepresenter
   property :label
   property :album
   property :year
-  property :excerpt_length, as: :length
+  property :excerpt_length, as: :duration
```
kookster: In `audio_version.rb` and `story.rb` models, the duration is defined at the model level, via alias or method.
I think for consistency it would be appropriate to alias `excerpt_length` to `duration` in the `musical_work.rb` model.

chrisrhoden: In both cases, there is additional work to be done and it's not just a straight database lookup.

In this case, `alias_method` will not work because there is no `excerpt_length` method until the first time it is accessed, meaning we would need to:
```ruby
def duration
  excerpt_length
end
```

`AudioFile` is set up to work the same way that this does – see [the representer](https://github.com/PRX/cms.prx.org/blob/b51f26b08e14888f7bc40a703cc537f605d9220e/app/representers/api/audio_file_representer.rb#L9). If this is wrong, then I'm not offering it as justification, but I can tell you that this was done this way for the same reasons.

This is high priority because it's currently breaking certain parts of beta. If you want me to add the method, I will do so, but I don't know that I agree it is more consistent.

kookster: it seems consistency is sorta out the window on this one. I am fine with it as is.

farski: FWIW `alias_attribute :duration, : excerpt_length` would have worked I think.

chrisrhoden: Learned a new method today, open a PR, dude!

PR #75, opened 2015-02-11
Alias duration for musical works in model

farski:
Moves the mapping for `excerpt_length` to `duration` to the model,
rather than the representer


chrisrhoden:
You should do this for `AudioFile` while you're having fun.

PR #78, opened 2015-04-02
Fix no method errors

kookster:
Seeing errors in production where getting the images, addresses, and stories related to accounts are throwing errors.

The controllers were still assuming these were has many rather than has one resources, and trying to retrieve and link to them incorrectly.

### DIFF
--- a/test/representers/api/account_representer_test.rb
+++ b/test/representers/api/account_representer_test.rb
@@ -6,9 +6,9 @@ require 'account' if !defined?(AudioFile)
 
 describe Api::AccountRepresenter do
 
-  let(:account)     { create(:account) }
+  let(:account) { create(:account) }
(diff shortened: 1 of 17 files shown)


test/representers/api/account_representer_test.rb
```diff
@@ -6,9 +6,9 @@
 
 describe Api::AccountRepresenter do
 
-  let(:account)     { build_stubbed(:account) }
+  let(:account) { create(:account) }
```
chrisrhoden: Why was this change necessary? Is the representer writing to the database?

kookster: to generate nested urls in the representers, we need the account id

chrisrhoden: rgr.

kookster:
@chrisrhoden anything else I should do for this?

PR #80, opened 2015-04-06
Add explicit value to AudioVersionRepresenter

chrisrhoden:
adds the value added in PRX/prx.org#121


kookster:
Should there be a validation on the audio_version to only allow those values to be set?

kookster:
@chrisrhoden if you think the validation is a non-issue, let me know and I'll go ahead and approve.

PR #81, opened 2015-04-06
Create audio file

kookster:
Adds the `upload` property which can be a URL to be used as the audio file's media asset.

Also add `set_account_uri` property to the audio_file so an audio file can be added to an account even if it is not yet added to a story.

Finally, some upgrades to gems for good measure.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -1,11 +1,11 @@
 source 'https://rubygems.org'
 
 # Bundle edge Rails instead: gem 'rails', github: 'rails/rails'
-gem 'rails', '~> 4.1.0'
+gem 'rails', '~> 4.1.10'
 
 # JWS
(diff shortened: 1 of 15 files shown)


chrisrhoden:
@kookster  When I post a URL as an upload, it looks like I have to do it with a publicly accessible HTTP uri, and cannot do `s3://prx-upload/...` style URIs (or at least, if that is supported it must have been added in the past and not in this PR)

How do you recommend I approach this?

kookster:
I thought you would want to support the http:// style - if that is not the case, I can add support for s3:// style.

chrisrhoden:
@kookster assigning back to you pending support for private s3:// style uris since we need it for current upload system

PR #83, opened 2015-04-13
Refactor controllers

kookster:
I noticed some inconsistencies in how controllers were overriding resource and resources.

### DIFF
--- a/app/controllers/api/addresses_controller.rb
+++ b/app/controllers/api/addresses_controller.rb
@@ -7,8 +7,7 @@ class Api::AddressesController < Api::BaseController
   filter_resources_by :account_id
 
   def resource
-    @address = account.address if !super && account
-    @address
+    @address ||= account.try(:address) || super
   end
 
   def account
--- a/app/controllers/api/base_controller.rb
+++ b/app/controllers/api/base_controller.rb
@@ -21,7 +21,7 @@ class Api::BaseController < ApplicationController
 
   cache_api_action :index
 
-  caches_action :entrypoint, cache_path: ->(c) { { _c: Api.version(api_version).cache_key } }
+  caches_action :entrypoint, cache_path: ->(_c) { { _c: Api.version(api_version).cache_key } }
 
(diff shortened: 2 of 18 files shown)


app/controllers/api/addresses_controller.rb
```diff
@@ -7,8 +7,7 @@ class Api::AddressesController < Api::BaseController
   filter_resources_by :account_id
 
   def resource
-    @address = account.address if !super && account
-    @address
+    @address ||= account.try(:address) || super
```
chrisrhoden: This is semantically different from the old behavior, but I don't know enough to know whether it matters. Just calling it out, but will trust if you tell me it's right.

app/controllers/api/base_controller.rb
```diff
@@ -46,7 +46,7 @@ def set_accepts
   end
 
   def user_not_authorized
-    render json: { error: 'You are not authorized to perform this action' },
-                   status: 401
+    message = { error: 'You are not authorized to perform this action' }
```
chrisrhoden: Worth making this a constant?

app/controllers/api/user_images_controller.rb
```diff
@@ -9,8 +9,7 @@ class Api::UserImagesController < Api::BaseController
   represent_with Api::ImageRepresenter
 
   def resource
-    @user_image = user.image if !super && user
-    @user_image
+    @user_image ||= user.try(:image) || super
```
chrisrhoden: Ditto

chrisrhoden:
@kookster just a couple of points, want to make sure things are all kosher before I merge.

app/controllers/api/addresses_controller.rb
kookster: I think it is more correct now. It previously checked `super` in the trailing if, but did not use it.

Now it correctly tries to get the address for an account_id if specified, and then falls back on the address id.  I suppose it could check in the opposite order, but as the routes and representers are set up to use addresses only as belonging to users and accounts which can only have one address, the account nested route is the default.

PR #86, opened 2015-04-13
Add missing user tests

kookster:


test/models/address_test.rb
```diff
@@ -2,9 +2,19 @@
 
 describe Address do
 
-  let(:address) { FactoryGirl.create(:address) }
+  let(:address) { create(:address) }
```
chrisrhoden: this is not exercised

kookster: true, I will remove it.

test/models/address_test.rb
```diff
@@ -2,9 +2,19 @@
...
 
   it 'has a table defined' do
     Address.table_name.must_equal 'addresses'
   end
+
+  it 'provides underlying account for user address' do
+    user.address.account.must_equal user.individual_account
+  end
+
+  it 'provides underlying account for user address' do
```
kookster: this is also a dupe and wrong desc, will update it.

kookster:
@chrisrhoden lgtm now.

PR #101, opened 2015-04-22
Add Authorization endpoints to API which lists accounts.

### LINKED ISSUE #99: Provide list of accounts for an autorized session
Basically just need a list of accounts in a way that lets us switch between them for the purposes of persisting to that account.

`/api/v1/authorization/accounts` seems like the way to go.

needed for wrapping PRX/www.prx.org#210 up in a nice bow

chrisrhoden:
resolves #99

### DIFF
--- a/app/controllers/api/authorizations_controller.rb
+++ b/app/controllers/api/authorizations_controller.rb
@@ -1,12 +1,13 @@
 # encoding: utf-8
 
 class Api::AuthorizationsController < Api::BaseController
+  private
   def resource
-    Authorization.new(User.find(prx_auth_token.user_id))
+    @auth ||= Authorization.new(user)
   end
 
   def resources
-    Authorization.new(User.find(prx_auth_token.user_id)).accounts
+    @accounts ||= resource.accounts
   end
 
   def index_collection
@@ -16,5 +17,9 @@ class Api::AuthorizationsController < Api::BaseController
       item_class: Account,
       item_decorator: Api::Min::AccountRepresenter
     )
+  end
+
+  def user
+    @user ||= User.find(prx_auth_token.user_id)
   end
 end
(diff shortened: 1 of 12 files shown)


chrisrhoden:
closes #99

app/controllers/api/authorizations_controller.rb
```diff
@@ -0,0 +1,20 @@
+# encoding: utf-8
+
+class Api::AuthorizationsController < Api::BaseController
+  def resource
+    Authorization.new(User.find(prx_auth_token.user_id))
```
kookster: I might DRY this out a bit and put `User.find(prx_auth_token.user_id)` into a find or get user method.

app/controllers/api/authorizations_controller.rb
```diff
@@ -0,0 +1,20 @@
...
+  end
+
+  def resources
+    Authorization.new(User.find(prx_auth_token.user_id)).accounts
+  end
+
+  def index_collection
```
kookster: I am reconciling myself to this, but it seems odd to have accounts and authorization resources returned by the same controller.

I can't think of a better option really; maybe making a new controller just for this nested `accounts` index action - but that seems like it is not worth the effort.

app/models/base_model.rb
```diff
@@ -4,18 +4,7 @@ class BaseModel < ActiveRecord::Base
...
-  attr_accessor :is_root_resource
-
-  def is_root_resource
-    !!@is_root_resource
-  end
-
-  def show_curies
-    is_root_resource
-  end
+  include RepresentedModel
```
kookster: Nice idea, good refactoring.

kookster:
@chrisrhoden I had only minor feedback on this, LGTM.
If you want to DRY up that User.find method, fine, if not, fine - I'll merge as is.

chrisrhoden:
@kookster let me know if I missed something

PR #102, opened 2015-04-24
add custom `#consume!` to honor request content_type

chrisrhoden:
Honor Content-Type for PUT and POST requests made to the API, which results in
actually writing to the represented models instead of failing silently.

https://github.com/apotonick/roar-rails/blob/a109b40/lib/roar/rails/controller_additions.rb#L27

This has been fixed in apotonick/roar-rails@e92bd1cb, so can be removed once the
gem is upgraded.

### DIFF
--- a/app/controllers/concerns/hal_actions.rb
+++ b/app/controllers/concerns/hal_actions.rb
@@ -16,14 +16,14 @@ module HalActions
   end
 
   def create
-    self.resource = consume!(resource)
+    consume_with_content_type! resource
     authorize resource
     resource.save!
     respond_with show_resource, show_options
   end
 
   def update
-    self.resource = consume!(resource)
+    consume_with_content_type! resource
     authorize resource
     resource.save!
     respond_with show_resource, show_options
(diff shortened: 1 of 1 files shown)


app/controllers/concerns/hal_actions.rb
```diff
@@ -139,6 +139,25 @@ def index_options
...
+  private
+
+  # TODO: Remove this method when we upgrade roar-rails
+  # Background: https://github.com/apotonick/roar-rails/blob/a109b40/lib/roar/rails/controller_additions.rb#L27
+  def consume_with_content_type!(model, options = {})
+    type = request.content_type
+    format = Mime::Type.lookup(type).try(:symbol)
+
+    if format.blank?
+      raise UnsupportedMediaType.new("Cannot consume unregistered media type '#{type.inspect}'")
```
chrisrhoden: Can't do much about it.

app/controllers/concerns/hal_actions.rb
```diff
@@ -16,14 +16,14 @@ def show
   end
 
   def create
-    self.resource = consume!(resource)
+    consume_with_content_type! resource
```
kookster: I'm surprised this works without the `self.resource` assignment, I remember needing that since subsequent calls assume that has been set here.

app/controllers/concerns/hal_actions.rb
kookster: This class is raising errors: `NameError: uninitialized constant HalActions::UnsupportedMediaType`

kookster: Looks like it got added here, perhaps not in this version of roar, or needs the module qualified name
`Roar::Rails::ControllerAdditions::UnsupportedMediaType` ?

https://github.com/apotonick/roar-rails/commit/144cdfabb55c81fdd25f97584bd2599c8a4ef64c#diff-60b9532947b2d4f7808021b77fc92b40R87

app/controllers/concerns/hal_actions.rb
chrisrhoden: I'm closing this PR because it's included in the refactor, but this works because `consume!` transforms the resource in place.

chrisrhoden: That is, you have to hand it an object (like `Story.new`) which it then modifies. `resource`, in this case, is our object which is modified.

PR #103, opened 2015-04-27
Refactor HalActions concern for better filtering and organization

chrisrhoden:
Break HalActions controller concern into multiple modules for easier navigation.

In `HalActions::Resources` module, change the way that `resources_base` and
`resources` behave so that `resources_base` can be used to build new resources
with the appropriate relationships based on route parameters.

Introduce the `scoped` and `included` relation transformers, which are skipped
when building models and generating a collection cache key.

Rename `with_params` to `filtered`
Rename `with_ordering` to `sorted`
Rename `with_paging` to `paged`

`resources_base` should now return a relation which can have the transformations
selectively applied to it depending on its use.

Action            | scoped | included | filtered | sorted | paged | find
----------------- | ------ | -------- | -------- | ------ | ----- | ----
`index`            | *yes*  | *yes*    | *yes*    | *yes*  | *yes* | no
`show`            | *yes*  | *yes*    | *yes*    | no     | no    | *yes*
`create`          | no     | no       | *yes*    | no     | no    | no
`update`          | *yes*  | *yes*    | *yes*    | no     | no    | *yes*
`index` cache key | *yes*  | no       | *yes*    | *yes*  | *yes* | no


chrisrhoden:
Woo! Finally. @kookster it is ready for your careful consideration.

app/controllers/api/accounts_controller.rb
```diff
@@ -6,11 +6,17 @@ class Api::AccountsController < Api::BaseController
 
   represent_with Api::AccountRepresenter
 
+  private
+
   def resources
-    @accounts ||= user.try(:accounts) || Account.active.member
+    @resources ||= super.member
```
kookster: This `#member` scope will filter out the individual account for a user, which previously would have been returned.
I don't know if that is correct, but it is a change.

app/controllers/api/accounts_controller.rb
```diff
@@ -18,6 +24,10 @@ def user
     @user ||= User.find(params[:user_id]) if params[:user_id]
   end
 
+  def scoped(relation)
+    relation.active
```
kookster: I like this better under scoped :+1:

app/controllers/api/audio_files_controller.rb
```diff
@@ -7,17 +7,17 @@ class Api::AudioFilesController < Api::BaseController
   filter_resources_by :audio_version_id
 
   # if the audio version is not set, and story is, set default av
-  def resource
+  def create_resource
     super.tap do |af|
-      if story && af.try(:new_record?)
+      if story
         af.audio_version_id ||= story.default_audio_version.id
         af.account_id ||= story.account_id
```
kookster: We talked abut about this as a validation - to make sure these are in sync.
We could also have the model do a `#before_validation` method to make sure the `audio_file.account_id` is set to the `audio_version.account_id` when the `audio_file.audio_version_id` is set.

app/controllers/api/accounts_controller.rb
chrisrhoden: Since the app doesn't really have any need to deal with users at this level
going forward I simplified this behavior.

app/controllers/api/promos_controller.rb
```diff
@@ -4,21 +4,12 @@ class Api::PromosController < Api::BaseController
 
   api_versions :v1
 
-  def resource
-    super.tap do |af|
-      if story && af.try(:new_record?)
-        af.audio_version_id ||= story.promos.id
-        af.account_id ||= story.account_id
```
kookster: So these are coming from the request params, provided by using a nested route to get to this call, right?

chrisrhoden: Basically, it's because there is no other way to get to the promos
controller than through a story, and the resources_base is what will be
built on. So instead of AudioFile.new, the controller is calling
story.promo_audio.build

app/controllers/api/audio_files_controller.rb
chrisrhoden: The rest of this is actually no longer necessary, it is handled by `resources_base` defined as below, so if we update `Story` to work that way we can get rid of this. I didn't want to overload the pull request.

app/controllers/api/accounts_controller.rb
kookster: :+1: fine with me.

app/controllers/api/audio_files_controller.rb
kookster: you mean update `Story` or `AudioFile`?

app/controllers/api/promos_controller.rb
```diff
@@ -4,21 +4,12 @@ class Api::PromosController < Api::BaseController
...
-
-  def resources
-    @promos ||= story.try(:promos_audio)
+  def resources_base
+    story.promos_audio
   end
 
   def story
-    @story ||= Story.find(params[:story_id]) if params[:story_id]
+    @story ||= Story.find(params[:story_id])
```
kookster: noting that this controller previously unnecessarily accommodated promos without a story specified.
this is not allowed by the routes, so this is an appropriate assumption.

app/controllers/api/audio_files_controller.rb
chrisrhoden: My mistake, yes, AudioFile

app/controllers/api/stories_controller.rb
```diff
@@ -11,16 +11,26 @@ def random
...
+      account.portfolio_stories
+    else
+      super
+    end
   end
 
-  def resources
-    @stories ||= resources_base.includes(
+  def filtered(resources)
+    if highlighted?
```
kookster: I'm confused why this `#highlighted?` check is necessary in both `#resources` and `#resources_base` - is this redundant, unnecessary, and redundant?

app/controllers/api/stories_controller.rb
```diff
@@ -11,16 +11,26 @@ def random
...
+      resources
+    else
+      super
+    end
+  end
+
+  def included(relation)
```
kookster: Is this going to be used for the `#show` cache key method? I think previously I tried not to do that.
Absolutely great that `#included` is excluded for the `index` cache key, but should it also be excluded for `show`?

app/controllers/concerns/hal_actions/actions.rb
```diff
@@ -0,0 +1,104 @@
+module HalActions::Actions
```
kookster: Can I just say these new modules are so much better now, thanks for doing this!

app/controllers/concerns/hal_actions/actions.rb
```diff
@@ -0,0 +1,104 @@
...
+  def index
+    respond_with index_collection, index_options
+  end
+
+  def show
+    respond_with show_resource, show_options
+  end
+
+  def create
+    create_resource.tap do |resource|
```
kookster: I would probably not use `resource` as the block arg name, just b/c of the confusion that we also have a `resource` method and so that could be slightly confusing to a less familiar reader, but obviously it is fine and safe thing to do.

app/controllers/concerns/hal_actions/actions.rb
```diff
@@ -0,0 +1,104 @@
...
+      resource.save!
+      respond_with root_resource(resource), show_options
+    end
+  end
+
+  def destroy
+    destroy_resource.tap do |resource|
+      authorize resource
+      resource.destroy
+      redirect_to destroy_redirect
```
kookster: I would not expect a redirect response for a successful API call to destroy, more likely a 20(0|2|4) response, right?

app/controllers/api/stories_controller.rb
chrisrhoden: It is necessary. Basically, when requesting portfolio stories we do not want to additionally apply the account_id based filter.

kookster: my bad, I was reading the diff wrong, and thought it was in `resources_base` and `resources`, not in `resources_base` and `filtered`. This makes sense.

app/controllers/concerns/hal_actions/actions.rb
```diff
@@ -0,0 +1,104 @@
...
+  end
+
+  def valid_params_for_action(action)
+    (params.slice(*self.class.valid_params_list(action)) || {}).tap do |p|
+      p[:zoom] = zoom_param if zoom_param
+    end
+  end
+
+  def zoom_param
+    @zoom_param ||= ->() do
```
kookster: I know I wrote this, but I wonder if it should just be regular `lambda` syntax for a multi-line lambda - do we have a style preference for that condition?

app/controllers/concerns/hal_actions/cache.rb
```diff
@@ -0,0 +1,34 @@
...
+    PagedCollection.new(
+      filtered(paged(sorted(scoped(resources_base)))),
+      request,
+      item_class: self.class.resource_class,
+      item_decorator: self.class.resource_representer
+    ).cache_key
+  end
+
+  def show_cache_path
+    show_resource.updated_at.utc.to_i
```
kookster: So this `show_resource` will have the `included` as part of the db request, yes? but all we need is the `updated_at` - perhaps overkill?

app/controllers/api/stories_controller.rb
chrisrhoden: I thought that the show cache key was pegged on the resource, which is only fetched once, is that not the case?

I can exclude it for `show` but it seems like it's sort of a toss up which is better. Roar is by far the most expensive part of our controller actions.

app/controllers/concerns/hal_actions/cache.rb
chrisrhoden: It is not clear to me what we gain by introducing more complexity here. As it works now, `show_resource` does not call the equivalent of `#included`, so we could remove it from `show_resource`. My inclination is to benchmark it, because I honestly couldn't tell you whether it slows things down to have the `include` or to have to make 2 separate resource calls on a cache miss.

app/controllers/concerns/hal_actions/actions.rb
chrisrhoden: 204 is probably most appropriate. will do.

app/controllers/api/stories_controller.rb
kookster: I concur, I think best to leave this for now since it is likely a rounding error compared with `roar` serialization.

chrisrhoden: #104 104

app/controllers/concerns/hal_actions/actions.rb
chrisrhoden: will fix

app/controllers/concerns/hal_actions/actions.rb
chrisrhoden: will do

app/controllers/concerns/hal_actions/cache.rb
chrisrhoden: #104

kookster:
I have no further comments, so I'll be happy to merge after the fixes are in.

chrisrhoden:
@kookster let me know if I missed something.

PR #106, opened 2015-05-04
Automatically mark stories as deleted and `v4`

### LINKED ISSUE #87: Hide v4 pieces from v3
Set app_version to v4, and deleted_at to now on create.

chrisrhoden:
Mark stories as deleted when created through cms.prx.org so that they do not
appear in v3.

Mark stories as `v4` so that we can show them in cms.prx.org even though they
are marked as deleted.

When a v3 story is deleted through cms.prx.org, mark as deleted.

When a v4 story is deleted through cms.prx.org, remove the record from the
database.

resolves #87


chrisrhoden:
Not sure if this should go to @farski instead. Feel weird about giving all of these to @kookster but I know that he's more likely to have concerns.

app/models/story.rb
```diff
@@ -27,6 +35,8 @@ class Story < BaseModel
   has_one :promos, -> { where(promos: true) }, class_name: 'AudioVersion', foreign_key: :piece_id
   has_one :license, foreign_key: :piece_id
 
+  after_initialize :set_app_version
```
kookster: pls consider using `before_validation(on: :create)`

PR #109, opened 2015-05-12
Event messages

kookster:
For communicating between services, for example, on publishing a new story.
Relies on the new `announce` gem for setting up SNS topics, SQS queues, and subscriptions, and for sending messages.  Integrates with ActiveJob and `shoryuken` for processing messages.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -44,7 +44,7 @@ gem 'oj_mimic_json'
 
 ## Messaging
 gem 'shoryuken'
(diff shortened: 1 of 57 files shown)


chrisrhoden:
@kookster looks pretty good, only thing is we should be making published_at read only in the story representer if we're adding these methods to write to it, yeah?

kookster:
yes, good point, will update the representer

kookster:
I should probably review all attrs for read-only - `id` for example

PR #113, opened 2015-06-03
Paged collection cache key

### LINKED ISSUE #108: Musical works paging behaving incorrectly
Eg: https://cms.prx.org/api/v1/stories/36935/musical_works.json?page=2

I would expect the `next` `_link` to be page 3, but currently it's returning page 2

kookster:
fix #108 

also, fix race condition in Digest::MD5 by using OpenSSL version which is thread safe

### DIFF
--- a/app/models/paged_collection.rb
+++ b/app/models/paged_collection.rb
@@ -1,6 +1,7 @@
 # encoding: utf-8
 
 require 'forwardable'
+require 'openssl'
 
 class PagedCollection
   extend ActiveModel::Naming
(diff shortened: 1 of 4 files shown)


app/models/paged_collection.rb
```diff
@@ -29,14 +30,12 @@ def initialize(items, request=nil, options=nil)
...
-    key_components << items.inject([0, 0]) do |keys, i|
-      keys[0] = keys[0] + i.try(:id).to_i.modulo(100)
-      keys[1] = [keys[1], i.try(:updated_at).try(:utc).to_i].max
-      keys
-    end.flatten.join("-")
-
+    item_keys = items.inject([]) do |keys, i|
+      keys << i.try(:id)
+      keys << i.try(:updated_at).try(:utc).to_i
+    end
```
chrisrhoden: Any reason we're building this up in memory rather than using the `Digest#update` method?

PR #114, opened 2015-06-09
Add web (alternate) link, profile link

kookster:
This refactors how common links are created into a concern instead of the base representer, and then uses those link methods in the base and other representers.


chrisrhoden:
@kookster want me to fix merge or do you want to?

kookster:
I'll merge

PR #116, opened 2015-06-16
Add original audio link to audio file

kookster:


app/models/audio_file.rb
```diff
@@ -116,6 +116,6 @@ def final_location?
   end
 
   def self.policy_class
-    AccountablePolicy
+    AudioFilePolicy
```
chrisrhoden: I think that this is no longer necessary since this is what Pundit will infer.

app/policies/audio_file_policy.rb
```diff
@@ -0,0 +1,5 @@
+class AudioFilePolicy < AccountablePolicy
+  def original?
+    update?
```
chrisrhoden: Do we want to start actually checking scopes at this point? Might be a good place to start - in this case I'd say we can do something like `token.authorized?(resource.account, 'read-private')`

app/models/audio_file.rb
kookster: I'll remove it.

test/controllers/api/audio_files_controller_test.rb
```diff
@@ -49,4 +49,20 @@
     get :index, api_request_opts(story_id: story.id)
     assert_response :success
   end
+
+  describe '#original' do
+
+    before { @controller.current_user = user }
```
chrisrhoden: I'd recommend rebasing against master to see how we need to do this now that tokens are being honored.

app/policies/audio_file_policy.rb
kookster: I don't have your changes yet for using tokens in this PR, should I merge them in, or does this work even without the changes in #115 ?

test/controllers/api/audio_files_controller_test.rb
kookster: I thought that was in #115 which isn't merged yet.

chrisrhoden: You're right, dumb mistake on my part

kookster:
@chrisrhoden this should be updated with master now.

app/policies/audio_file_policy.rb
kookster: ok, where can I see scopes I can use?

kookster: https://github.com/PRX/id.prx.org/blob/master/app/models/authorization.rb#L4

kookster:
@chrisrhoden ping

PR #120, opened 2015-07-27
Add Series links to Account representers

farski:


app/representers/api/min/account_representer.rb
```diff
@@ -39,4 +39,12 @@ class Api::Min::AccountRepresenter < Api::BaseRepresenter
     }
   end
   embed :stories, paged: true, item_class: Story, item_decorator: Api::Min::StoryRepresenter, zoom: false
+
+  link :series do
+    {
+      href: "#{api_account_series_index_path(represented)}{?filters}",
```
kookster: Not sure why I made the `min` version of this link not match the regular representer, but I think they probably should both include paging (as should the link to pieces in the min).

app/representers/api/min/account_representer.rb
```diff
@@ -34,9 +34,17 @@ class Api::Min::AccountRepresenter < Api::BaseRepresenter
 
   link :stories do
     {
-      href: "#{api_account_stories_path(represented)}{?filters}",
+      href: "#{api_account_stories_path(represented)}{?page,per,zoom,filters}",
```
farski: @kookster Something like this?

kookster: yes, exactly.

PR #121, opened 2015-08-03
Update bundled gems

farski:


chrisrhoden:
@farski looks great thanks! you should deploy with this merged.

PR #128, opened 2016-02-26
Fix cross origin requests

cavis:
Adding `headers: :any` to allow non-whitelisted domains to do the preflight `Access-Control-Request-Method: GET`.

Also whitelisted the `prx.tech` domain.

### DIFF
--- a/config/application.rb
+++ b/config/application.rb
@@ -45,7 +45,7 @@ module PRX
 
       allow do
         origins '*'
-        resource '/api/*', methods: [:get]
+        resource '/api/*', methods: [:get], headers: :any
       end
     end
 
--- a/config/application.rb
+++ b/config/application.rb
@@ -39,7 +39,7 @@ module PRX
 
     config.middleware.insert_after Rails::Rack::Logger, Rack::Cors do
       allow do
-        origins /.*\.prx\.(?:org|dev)$/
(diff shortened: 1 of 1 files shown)


config/application.rb
```diff
@@ -39,13 +39,13 @@ class Application < Rails::Application
...
       allow do
-        origins /.*\.prx\.(?:org|dev)$/
+        origins /.*\.prx\.(?:org|dev|tech)$/
         resource '/api/*', methods: [:get, :put, :post, :options], headers: :any
       end
 
       allow do
         origins '*'
-        resource '/api/*', methods: [:get]
+        resource '/api/*', methods: [:get], headers: :any
```
chrisrhoden: I can't merge this – we actually don't know what the implications of allowing anyone to make authenticated requests against the API are. We really just want to let them make unauthenticated public requests.

cavis: So except for unauthorized GET requests, we're keeping the CMS api internal-only for the time being?  That's fine - just want to be clear.

chrisrhoden: For now, yeah. I mean, there's not really a way for anyone else to get tokens at this point anyway, so it's sort of moot. Just don't want to open this up until it's necessary.

cavis:
There we go!  Lines changed: 1.

PR #133, opened 2016-04-07
Authorized endpoints

cavis:
Adds some more authorized endpoints.

- `GET /authorization` the current user
- `GET /authorization/accounts` active and approved accounts of the user
- `GET /authorization/accounts/:id` show account of the user by id
- `GET /authorization/accounts/:id/stories` list ALL (including unpublished) stories of the account
- `POST /authorization/accounts/:id/stories` create new story for the account
- `GET /authorization/stories/:id` show any story of any of the users accounts
- `DELETE /authorization/stories/:id` destroy any story


.hound.yml
```diff
@@ -1,3 +1,3 @@
 ruby:
   enabled: true
-  config_file: config/.rubocop.yml
+  config_file: .rubocop.yml
```
cavis: Moved rubocop to the top-level, so it gets auto-detected locally.

app/controllers/concerns/hal_actions/actions.rb
```diff
@@ -42,6 +42,8 @@ def destroy_redirect
   def index_options
     valid_params_for_action(:index).tap do |options|
       options[:_keys] = options.keys
+      # TODO: this isn't great
+      options[:represent_with] = Api::PagedCollectionRepresenter
```
cavis: I _think_ this is okay, as long as we're always going to represent `index` actions with a paged collection.

kookster: For sure, that is the plan, and how I would expect things to work to maintain consistency. Now, if it ceases to be the plan, a single controller could overwrite this method, or we could modify when we get there.

app/controllers/concerns/hal_actions/errors.rb
```diff
@@ -13,6 +13,16 @@ def status
...
+  class NotFound < StandardError
+    def message
+      'Resource not found'
+    end
+
+    def status
+      404
+    end
+  end
+
```
kookster: yeah, we added this kind of not found handling to the extracted gem. This is fine for now, will have to refactor when we pull this in: https://github.com/PRX/hal_api-rails/pull/1/files

cavis: Sounds good.  I added this to catch some activerecord-not-found errors in very specific places, for the show/update/destroy actions.  Hopefully specific enough that other unrelated `model.find()` errors won't trigger an incorrect 404.

app/representers/api/authorization_representer.rb
```diff
@@ -18,9 +19,12 @@ class Api::AuthorizationRepresenter < Api::BaseRepresenter
       count: represented.accounts.count
     }
   end
-  embed :accounts, paged: true, item_class: Account, item_decorator: Api::Min::AccountRepresenter, per: :all
+  embed :accounts, get: :approved_active_accounts, paged: true, per: :all,
+                   item_class: Account, item_decorator: Api::Auth::AccountMinRepresenter,
+                   url: lambda { |_r| api_authorization_accounts_path }
```
kookster: do we need to add `:get` here?
could you do `embed :approved_active_accounts, as: :accounts`?

like this one:
`embeds :audio_files, as: :audio, class: AudioFile, decorator: Api::AudioFileRepresenter`
?

app/representers/concerns/embeds.rb
```diff
@@ -63,14 +63,14 @@ def embed(name, options={})
...
         }
+        getter_per = options.delete(:per) || Kaminari.config.default_per_page
+        getter_name = options.delete(:get) || name
         options[:getter] ||= ->(*) do
           # set # per page based on default, option value integer, or special :all
-          per = opts.delete(:per)
-          per = self.send(name).count if per == :all
-          PagedCollection.new(self.send(name).page(1).per(per), nil, opts.merge({parent: self}))
+          per = send(getter_name).count if getter_per == :all
+          PagedCollection.new(send(getter_name).page(1).per(per), nil, opts.merge(parent: self))
```
kookster: see above.
seems like you are trying to have the `name` be one thing, and the `getter` another, which is fine, but I think the `:as` option already does this.

kookster:
@cavis I think that's it for my review - the `:get` vs. `:as` use is about all I see that could be changed (assuming I am right that :as can do what you need)

app/representers/api/authorization_representer.rb
cavis: Ah!  I didn't know such a thing existed.  Will update, and remove my changes to `embeds.rb`

app/representers/concerns/embeds.rb
cavis: Will get rid of the `:get` option.  But do you agree that this shouldn't be `opts.delete(:per)`-ing inside the getter?  I think that would work normally on the first call, but on the second call you would get `.page(1).per(nil)`.

kookster: I believe you, probably an oversight on my part, thanks for fixing

PR #137, opened 2016-04-20
Update to use the `hal_api-rails` gem

kookster:


app/representers/api/base_representer.rb
```diff
@@ -1,25 +1,20 @@
...
       name: :prx,
-      href: "http://#{prx_meta_host}/relation/{rel}",
+      href: "http://#{profile_host}/relation/{rel}",
       templated: true
     }]
   end
 
-  self_link
+  def self.alternate_host
+    "www.prx.org"
```
cavis: isn't this an ENV variable?

app/representers/api/base_representer.rb
```diff
@@ -1,25 +1,20 @@
...
+  end
 
-  profile_link
+  def self.profile_host
+    "meta.prx.org"
```
cavis: Same here ... ENV?

app/representers/api/auth/account_representer.rb
```diff
@@ -1,14 +1,13 @@
...
 class Api::Auth::AccountRepresenter < Api::AccountRepresenter
-
   # point to authorized stories (including unpublished)
   link :stories do
     {
       href: "#{api_authorization_account_stories_path(represented)}{?page,per,zoom,filters}",
       templated: true,
       count: represented.all_stories.count
-    }
+    } if represented.id
```
cavis: What is this doing?  Only displaying the link at `/accounts/:id`, vs on `/accounts`?

config/initializers/mime_types.rb
```diff
@@ -3,11 +3,3 @@
 # Add new mime types for use in respond_to blocks:
 # Mime::Type.register "text/richtext", :rtf
 # Mime::Type.register_alias "text/html", :iphone
```
cavis: Is this initializer needed anymore?  Or just keeping it here for posterity?

cavis:
Branch seems to "mostly" work against my local publish.prx.org - but that's because it's a few commits behind master.  Generally looks good to me.

But what about houndci-bot?

app/representers/api/auth/account_representer.rb
kookster: this if there is a render of the account when it has not yet been persisted and so there is no id for it yet.

PR #138, opened 2016-04-27
Audio file fields

cavis:
Updates for publish.prx.org file uploading:

- Add the writeable `position`
- Add the read-only `status`
- Infer `account_id` through the `audio_version.story.account_id`, if possible.  So when POSTing new files through the `/audio_versions/:id/audio_files`, they get that implicit account set.

### DIFF
--- a/app/representers/api/audio_file_representer.rb
+++ b/app/representers/api/audio_file_representer.rb
@@ -3,8 +3,11 @@
 class Api::AudioFileRepresenter < Api::BaseRepresenter
 
   property :id, writeable: false
+  property :position
+
   property :filename
   property :label
+  property :status, writeable: false
   property :size
   property :duration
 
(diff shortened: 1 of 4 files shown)


app/representers/api/audio_file_representer.rb
```diff
@@ -3,8 +3,11 @@
 class Api::AudioFileRepresenter < Api::BaseRepresenter
 
   property :id, writeable: false
+  property :position
+
   property :filename
   property :label
+  property :status, writeable: false
```
cavis: I assume we'd never want to set this through the CMS, right?

test/test_helper.rb
```diff
@@ -6,7 +6,7 @@
 
 require 'simplecov' if !ENV['GUARD'] || ENV['GUARD_COVERAGE']
 
-ENV['action_dispatch.logger'] = Rails.logger
+ENV['action_dispatch.logger'] = Rails.logger if defined? Rails
```
cavis: I sometimes run tests like `ruby -I test/ test/controllers/api/audio_files_controller_test.rb`

app/representers/api/audio_file_representer.rb
kookster: right.

PR #139, opened 2016-05-03
Fix image profile uris

kookster:
Mostly involves upgrading the hal api rails gem

### DIFF
--- a/test/test_helper.rb
+++ b/test/test_helper.rb
@@ -6,8 +6,6 @@ ENV['META_HOST'] = 'meta.prx.org'
 
 require 'simplecov' if !ENV['GUARD'] || ENV['GUARD_COVERAGE']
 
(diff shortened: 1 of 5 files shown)


test/test_helper.rb
```diff
@@ -6,8 +6,6 @@
 
 require 'simplecov' if !ENV['GUARD'] || ENV['GUARD_COVERAGE']
 
-ENV['action_dispatch.logger'] = Rails.logger if defined? Rails
```
cavis: Throwing errors?  I thought the `if defined?` prevented that?

kookster: The error is b/c you can't set an env var value to an object, only a string:
```
[1] pry(main)> ENV['action_dispatch.logger'] = Rails.logger
TypeError: no implicit conversion of ActiveSupport::Logger into String
```

cavis:
That'll do, @kookster.  That'll do.

PR #141, opened 2016-05-13
Cache busting

### LINKED ISSUE #135: Tags over-cached on update requests
Making a PUT request to a story with `{"tags":["Foobar"]}` doesn't seem to bust the cache in production.  So subsequent GET requests will come back with the un-altered tags, until the cache expires some minutes later.

Should re-check this after @kookster's latest PR, and investigate if that doesn't fix it.

cavis:
Brute force method: add `touch: true` to every `belongs_to` association.

(Except for the user-default-account ... there's a circular relation there).

Hopefully fixes #135.

### DIFF
--- a/app/models/story.rb
+++ b/app/models/story.rb
@@ -14,9 +14,9 @@ class Story < BaseModel
     )
   end
 
-  belongs_to :account, -> { with_deleted }
-  belongs_to :creator, -> { with_deleted }, class_name: 'User', foreign_key: 'creator_id'
-  belongs_to :series
+  belongs_to :account, -> { with_deleted }, touch: true
+  belongs_to :creator, -> { with_deleted }, class_name: 'User', foreign_key: 'creator_id', touch: true
+  belongs_to :series, touch: true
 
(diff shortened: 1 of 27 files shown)


cavis:
Confirmed that this fixes my problems (running locally w/caching enabled).

app/models/story.rb
```diff
@@ -14,10 +14,9 @@ def self.paranoia_scope
     )
   end
 
-  belongs_to :account, -> { with_deleted }, touch: true
-  belongs_to :creator, -> { with_deleted }, class_name: 'User', foreign_key: 'creator_id',
-                                            touch: true
-  belongs_to :series, touch: true
+  belongs_to :account, -> { with_deleted }
```
cavis: When you get an account under the `/authorization` endpoint, aren't stories embedded in it?

app/models/story.rb
```diff
@@ -14,10 +14,9 @@ def self.paranoia_scope
...
+  belongs_to :creator, -> { with_deleted }, class_name: 'User', foreign_key: 'creator_id'
+  belongs_to :series
```
cavis: Similarly, could embedded stories-in-a-series get over cached?

PR #142, opened 2016-05-19
Dockerize

cavis:
Setup to run in ECS staging, but still prod-deployable via capistrano.


cavis:
Okay, think this is good to go.

May need to revisit getting codecov to comment on PRs ... I [see the reports there](https://codecov.io/gh/PRX/cms.prx.org), but it's not commenting in this thread.  May just be a setting.

Dockerfile
```diff
@@ -0,0 +1,41 @@
+FROM alpine:3.3
+
+MAINTAINER PRX <sysadmin@prx.org>
+
+RUN echo "http://nl.alpinelinux.org/alpine/edge/testing" >> /etc/apk/repositories && \
```
kookster: I think 3.4 is out now, is there something in edge we need?

cavis: Edge is the only place to get `autossh`.  This branch still supports ssh-tunneling mysql, even though that doesn't seem viable at this point.  Shall I gut that part, for now?  Or hold out hope that it'll get faster someday?

kookster: I think gut it, and we'll add a version of the db to the staging rds.

cavis: Will do.  And I've already added a staging RDS mysql instance, hooked up in ECS -> http://staging.cms.prx.tech

Much faster than tunneling out.

bin/application
```diff
@@ -0,0 +1,64 @@
...
+  echo "usage: application [ help | web | migrate | test | -- ]"
+}
+
+CmsParseOpts () {
+  PROCESS=$1
+  shift
+  CMD_ARGS=$*
+}
+
+CmsTunnelMysql () {
```
kookster: removing this too?

PR #143, opened 2016-07-11
Support for story images

cavis:
Backend support for publishing story images.

- Misc sorting/filtering options for dashboard (list stories-in-series, list newest-series, list stories-without-series, etc)
- Fix prying (byebug-navigation and less support)
- Abstract the background-processing of audio files into the `Fixerable` concern.  Nothing really changes here, but paves the way for background-processing of image files.
- Temporary support for setting `remote_file_url` via the image `upload` param.  And imagemagick-ing image thumbnails directly.  (Requires write s3 credentials ... which we may not really want to do).

Note: will need to update staging CMS to have its own non-production bucket.  And maybe copy over some of the prod-bucket image folders, so things don't look completely broken.


Guardfile
```diff
@@ -3,7 +3,7 @@
 
 notification :growl
 
-guard :minitest, spring: true, env: {GUARD: 'true'}, all_env: :GUARD_COVERAGE, all_after_pass: true do
+guard :minitest, spring: true, env: {GUARD: 'true'}, all_env: :GUARD_COVERAGE, all_after_pass: false do
```
kookster: just too slow? I usually like having all run after changes, but understandable to change this.

app/controllers/api/auth/series_controller.rb
```diff
@@ -0,0 +1,21 @@
...
+  include ApiAuthenticated
+
+  api_versions :v1
+
+  filter_resources_by :account_id
+
+  represent_with Api::Auth::SeriesRepresenter
+
+  def sorted(res)
+    res.order('updated_at desc')
```
kookster: this is inconsistent with the non-auth series controller, perhaps we should make them match?

app/controllers/api/auth/stories_controller.rb
```diff
@@ -6,19 +6,44 @@ class Api::Auth::StoriesController < Api::StoriesController
...
+  def create_resource
+    super.tap do |story|
+      story.creator_id = current_user.id
+      story.account_id ||= story.series.try(:account_id)
+      story.account_id ||= current_user.account_id
+      story.account_id ||= current_user.approved_accounts.first.try(:id)
+    end
+  end
+
+  def filters
```
kookster: Is this worth lifting up to a mixin / superclass? I see it in stories and series controllers, getting to be a regular pattern.

app/models/audio_file.rb
```diff
@@ -3,6 +3,7 @@
 class AudioFile < BaseModel
 
   include PublicAsset
+  include Fixerable
```
kookster: such naming!

app/representers/api/auth/series_representer.rb
```diff
@@ -0,0 +1,25 @@
...
+      count: represented.all_stories.count
+    } if represented.id
+  end
+  embed :all_stories, as: :stories, paged: true, item_class: Story,
+                      item_decorator: Api::Auth::StoryMinRepresenter,
+                      url: ->(_r) { api_authorization_series_stories_path(represented.parent) },
+                      zoom: false
+
+  # TODO: why is this necessary? maybe because :all_stories != :stories ???
+  embed :stories, zoom: false
```
kookster: seems like we don't need this line.

app/representers/api/image_representer.rb
```diff
@@ -2,12 +2,30 @@
...
+  def self_url(represented)
+    if represented.is_a?(StoryImage)
+      api_story_story_image_path(represented.story, represented)
+    elsif represented.is_a?(SeriesImage)
+      api_series_series_image_path(represented.series, represented)
+    elsif represented.is_a?(AccountImage)
+      api_account_account_image_path(represented.account, represented)
+    elsif represented.is_a?(UserImage)
+      api_user_user_image_path(represented.user)
+    else
```
kookster: seems like there should be a way to meta program this, using the class name, but this is safer probably.

Guardfile
cavis: i may not have intended to check that in. but yes - this particular repo is pretty slow. especially when testing through docker. makes it a pain to fix a particular test if i have to wait minute(s) to see the results.

and you can always hit enter to run all the tests again, right?

app/representers/api/image_representer.rb
```diff
@@ -2,12 +2,30 @@
...
   property :id, writeable: false
-  property :filename
+  property :filename, writeable: false
   property :size
   property :caption
   property :credit
 
+  # provide either an accessible url or the file itself for upload
+  property :upload, readable: false
+  property :file, readable: false
```
kookster: I don't understand why we need to show both of these - I thought these were carrierwave managed attributes, but perhaps I am confused.

app/controllers/api/auth/series_controller.rb
cavis: This was just a fast way to get the sorted-ness I wanted.  But I'm willing to debate how we handle sorting.  Is there any existing query param?

app/controllers/api/auth/stories_controller.rb
cavis: Yeah, I think it's probably worth standardizing on both filters and sorting.

app/models/audio_file.rb
cavis: I have a gift.

kookster:
@cavis a few comments, a few questions, overall looks solid.

app/representers/api/auth/series_representer.rb
cavis: Yeah, this is definitely working around a bug.  (That all_stories `embed` doesn't respect its own `zoom: false`).  But if we're not into using the `?zoom` param anyways, I can probably just take out the embed.

Although I may need something like this line, to override the `embed` from the parent representer.

app/representers/api/image_representer.rb
cavis: Yeah, I started down that road.  But the singular `UserImage` (where everything else is plural) makes that a bit odd.  So I opted for explicit-everything.

app/representers/api/image_representer.rb
cavis: This will change with background processing.

But: `upload` (named to match the existing audio-file `upload` param) is currently aliased to `remote_file_url`.  So you give it an s3 url, it downloads the file, and sets `image.file = downloaded_file`.  When we switch over to background processing, the param name can probably stay the same.  But instead of CMS downloading the file, it'll just wait for the lambda to finish working.

I left the `file` param, in case I wanted to directly POST the file to cms.  Seems like we never will, though.

PR #144, opened 2016-07-12
Networks

kookster:
Include new endpoints for network resources, with authenticated endpoints to return all stories in a network (including the stories only visible to members of the network where `network_only_at` will be not null).

Add the `transcript` field to the full story json, and to audio versions.


app/controllers/api/auth/stories_controller.rb
```diff
@@ -32,7 +38,12 @@ def filtered(resources)
   end
 
   def resources_base
-    @stories ||= current_user.approved_account_stories
+    # If there is a network_id specified, use that network
```
cavis: This implies that the "network-visible stories" are always going to be a subset of `current_user.approved_account_stories`... right?  Or am I misreading that?

app/controllers/api/stories_controller.rb
```diff
@@ -68,7 +68,7 @@ def included(relation)
   end
 
   def scoped(relation)
-    relation.published.visible
+    relation.published.network_visible.series_visible
```
cavis: Can you explain this?  I'm not sure I understand what both those statements are doing, assuming the user is supplying only one of `network_id` or `series_id`.

app/controllers/api/auth/stories_controller.rb
kookster: see line 43, where if there is a network id, then the network visible stories are a subset of `super.published`

app/models/story.rb
```diff
@@ -52,7 +53,7 @@ def self.paranoia_scope
   # indicates the piece is not publically available, only to the network
   event_attribute :network_only_at
 
-  scope :published, -> { where('`published_at` IS NOT NULL AND `network_only_at` IS NULL') }
+  scope :published, -> { where('`published_at` IS NOT NULL') }
```
cavis: Oh, I see, you changed what the `published` scope means.

app/models/story.rb
```diff
@@ -117,6 +120,14 @@ def tags=(ts)
     self.user_tags = ts.uniq.sort.map { |t| UserTag.new(name: t) }
   end
 
+  def transcript
```
cavis: What is this story/audio_file transcript supporting?  Just a coincidental feature?

app/controllers/api/stories_controller.rb
kookster: sure. Previously the `visible` scope was really doing 2 things: it filtered out pieces that were `network_only`, or were in a series where the pieces were `subscriber_only`. I made the two scopes for `network_visible` and `series_visible` from the single `visible` scope, because there are conditions when I want to use one or the other but not both.

cavis:
Looks good to me.  👍

app/models/story.rb
kookster: It is another request from the AMPERs folks for their network integration. I decided just to put it in with the other networks related changes for them.

PR #146, opened 2016-07-26
Async image processing

cavis:
- Use `upload_path` and `status` to create pending uploads
- Announce image changes to lambda worker via sns
  - Creates and updates will cause the worker to resize the image, and move to the destination bucket/path.
  - Deletes will hard-delete the mysql record, but the worker will NOT delete the file, instead putting a `deleted.json` file representation of the last state of the record.
- Get callbacks from workers via shoryuken
  - Updates image name/status/size/etc based on the imagemagick identify output

### DIFF
--- a/app/controllers/concerns/announce_actions/announce_filter.rb
+++ b/app/controllers/concerns/announce_actions/announce_filter.rb
@@ -42,7 +42,6 @@ module AnnounceActions
     def decorator_class(controller)
       return options[:decorator] if options[:decorator]
       resource_class = controller.controller_name.singularize.camelize
-      "Api::Msg::#{resource_class}Representer".safe_constantize ||
       "Api::Min::#{resource_class}Representer".safe_constantize ||
         "Api::#{resource_class}Representer".safe_constantize
     end
--- a/app/controllers/concerns/announce_actions/announce_filter.rb
+++ b/app/controllers/concerns/announce_actions/announce_filter.rb
@@ -42,6 +42,7 @@ module AnnounceActions
     def decorator_class(controller)
       return options[:decorator] if options[:decorator]
       resource_class = controller.controller_name.singularize.camelize
+      "Api::Msg::#{resource_class}Representer".safe_constantize ||
       "Api::Min::#{resource_class}Representer".safe_constantize ||
         "Api::#{resource_class}Representer".safe_constantize
     end
--- a/app/controllers/concerns/announce_actions/announce_filter.rb
+++ b/app/controllers/concerns/announce_actions/announce_filter.rb
@@ -16,8 +16,11 @@ module AnnounceActions
     end
 
     def after(controller)
-      subject = controller.controller_name.singularize
+      subject = (options[:subject] || controller.controller_name).to_s.singularize
       announce(subject, resource_action, decorated_resource(controller))
+    rescue Aws::SNS::Errors::NotFound => e
+      Rails.logger.error("Non-existent SNS topic: #{Rails.env.downcase}_announce_#{subject}_#{action}")
+      raise e
     end
 
     def resource_action
--- a/app/models/audio_file.rb
+++ b/app/models/audio_file.rb
@@ -26,6 +26,7 @@ class AudioFile < BaseModel
   alias_attribute :duration, :length
 
   mount_uploader :file, AudioFileUploader, mount_on: :filename
+  skip_callback :commit, :after, :remove_file! # don't remove s3 file
   fixerable_upload :upload, :file
 
   before_validation do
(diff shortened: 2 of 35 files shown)


app/controllers/concerns/announce_actions/announce_filter.rb
```diff
@@ -39,6 +42,7 @@ def announce_resource(action, controller)
     def decorator_class(controller)
       return options[:decorator] if options[:decorator]
       resource_class = controller.controller_name.singularize.camelize
+      "Api::Msg::#{resource_class}Representer".safe_constantize ||
```
kookster: Is this common enough to need to be here, instead of just using the `decorator` option?

app/models/audio_file.rb
```diff
@@ -26,6 +26,7 @@ class AudioFile < BaseModel
   alias_attribute :duration, :length
 
   mount_uploader :file, AudioFileUploader, mount_on: :filename
+  skip_callback :commit, :after, :remove_file! # don't remove s3 file
```
kookster: or https://github.com/carrierwaveuploader/carrierwave/wiki/How-to%3a-Keep-uploaded-files-%28for-a-specific-uploader%29

not sure which method is better, the `config.remove_previously_stored_files_after_update = false` seems more directly meant for this.

app/controllers/concerns/announce_actions/announce_filter.rb
cavis: Yeah, it's probably going to be just audio and images.  The explicit would probably be better in this case.  Will update!

app/models/image.rb
```diff
@@ -5,9 +5,25 @@ class Image < BaseModel
...
+  NOTFOUND = 'not found'.freeze
+  INVALID  = 'invalid'.freeze
+  FAILED   = 'failed'.freeze
+  COMPLETE = 'complete'.freeze
+
+  alias_attribute :upload, :upload_path
 
   mount_uploader :file, ImageUploader, mount_on: :filename
-  alias_attribute :upload, :remote_file_url
+  skip_callback :commit, :after, :remove_file! # don't remove s3 file
```
kookster: same comment as above - and since we generally always want to keep files, I think maybe a global config makes more sense.

app/models/audio_file.rb
cavis: I think it was [this post](http://stackoverflow.com/questions/17302678/i-would-like-to-stop-carrierwave-deleting-a-file) that led me to believe that only applies to updates.  Will have to double-check that - never actually tried that config.

cavis: @kookster we actually already set `config.remove_previously_stored_files_after_update = false` in the carrierwave initializer.

But I confirmed that setting only applies to updates, not deletes.  Without the `skip_callback` the files will all be removed from S3, assuming CMS has permissions to do so.  (If it doesn't, you'll get a 500).

cavis:
@kookster couple more changes to support audio files.

Not exactly sure how to get the "Channel Mode" from ffmpeg output ... but can revisit that later.  For now, it just sets it to single channel or stereo.

Also added some more statuses to `audio_files`.  Does that make sense, or would I be better off sticking to the existing `creating mp3s` and `creating mp3s failed` statuses?

app/models/audio_file.rb
kookster: my bad, thanks for checking!

cavis:
@kookster ping!

app/models/image.rb
kookster: and like above, nvm

app/representers/api/image_representer.rb
```diff
@@ -17,19 +17,26 @@ def self_url(represented)
   end
 
   property :id, writeable: false
-  property :filename, writeable: false
+  property :filename
```
kookster: I thought carrierwave would prevent direct update of this attribute, why this change?

cavis: This was part of my initial pass, where CMS itself (carrierwave) was doing the resize/thumbnails.  Mostly it was useful in dev to leave S3/Lambdas out.

But now that we've moved away from that, I can see removing direct-PUT-ing the file.  Should also update the Dockerfile to not include imagemagick, then.

config/environments/development.rb
```diff
@@ -27,6 +27,9 @@
   # Raise an error on page load if there are pending migrations
   config.active_record.migration_error = :page_load
 
+  # Better logging for docker-compose
+  config.logger = Logger.new(STDOUT)
```
kookster: yeah, I think this should become standard, I was doing something similar on feeder, but this is a better way.

kookster:
whew! 👍

PR #150, opened 2016-09-06
Helper task to load prod database

cavis:
Some rake tasks to dump/scp/restore a database to your dev environment.  Makes getting test data into your docker-compose mysql much easier.


kookster:
did a merge with master so this should be good to merge.

PR #151, opened 2016-09-06
Simple forever-based worker for production server

### LINKED ISSUE #149: devops web/worker setup
After #146 we need a worker process in addition to the web.

- [x] Add a worker service to ECS staging
- [x] Configure snap pipeline to deploy staging web+worker
- [ ] Setup a worker on production (capistrano)

cavis:
Added `forever` to production.  Simple cap task to start/restart the `bin/application worker` script.

Fixes #149.

### DIFF
--- a/config/deploy.rb
+++ b/config/deploy.rb
@@ -52,7 +52,6 @@ namespace :deploy do
     on roles(:app), in: :sequence, wait: 5 do
       within release_path do
         with rails_env: fetch(:rails_env) do
-          cmd = release_path.join('bin/application')
           log = release_path.join('log/production.worker.log')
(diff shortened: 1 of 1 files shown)


kookster:
we have `god` running on prx01 and prx04 for the pollers, perhaps we should use that for another worker instead, since it's already there?
(expected answer: now you tell me!)

cavis:
Whoops - didn't even realize there were pollers on prx04!  (Saw them on prx01).  Probably worth revisiting this with `god` instead of `forever`.

That said ... the `forever` worker is actually already running on prx04.  So I'm not in a huge hurry.

PR #152, opened 2016-09-09
Series/story v4 filtering

cavis:
Ability to filter stories/series based on their `app_version`.

Part of PRX/publish.prx.org#27

Requires migration PRX/prx.org#282 (and copying that to the staging DB).

### DIFF
--- a/app/controllers/api/series_controller.rb
+++ b/app/controllers/api/series_controller.rb
@@ -5,4 +5,11 @@ class Api::SeriesController < Api::BaseController
   api_versions :v1
 
   filter_resources_by :account_id
+
+  filter_params :v4
+
+  def filtered(resources)
+    resources = resources.v4 if filters.v4?
+    super
+  end
 end
--- a/app/models/image.rb
+++ b/app/models/image.rb
@@ -25,15 +25,6 @@ class Image < BaseModel
     end
   end
 
-  # not all the tables have these columns, story images do
-  def caption
-    self.try(:read_attribute, :caption)
-  end
-
-  def credit
-    self.try(:read_attribute, :credit)
-  end
-
   def self.policy_class
     ImagePolicy
   end
--- a/app/models/series.rb
+++ b/app/models/series.rb
@@ -19,7 +19,11 @@ class Series < BaseModel
 
   has_one :image, -> { where(parent_id: nil) }, class_name: 'SeriesImage'
 
+  before_validation :set_app_version, on: :create
+
   event_attribute :subscriber_only_at
+
+  scope :v4, -> { where('`app_version` = ?', PRX::APP_VERSION) }
 
   def story_count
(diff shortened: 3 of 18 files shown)


test/controllers/api/auth/series_controller_test.rb
```diff
@@ -0,0 +1,57 @@
+require 'test_helper'
```
cavis: Not sure why hound is picking this up - I don't get it from rubocop locally.

app/models/image.rb
```diff
@@ -25,15 +25,6 @@ class Image < BaseModel
...
 
-  # not all the tables have these columns, story images do
-  def caption
-    self.try(:read_attribute, :caption)
-  end
-
-  def credit
-    self.try(:read_attribute, :credit)
-  end
-
```
kookster: removing this b/c they all will have now these columns now, so no need for these accessors.

app/models/series.rb
```diff
@@ -104,4 +108,23 @@ def schedule_for_episode_number(episode_number)
 
     schedule_index
   end
+
+  def destroy
+    if v4?
+      really_destroy!
```
kookster: I thought we said we were treating v4 series like v3 - no need for using paranoid to filter them out?

app/controllers/api/series_controller.rb
```diff
@@ -5,4 +5,11 @@ class Api::SeriesController < Api::BaseController
   api_versions :v1
 
   filter_resources_by :account_id
+
+  filter_params :v4
```
kookster: filter makes sense to me since we have the API supporting beta which shows both v3 and v4. sensible.

app/models/image.rb
cavis: correct

app/models/series.rb
cavis: This doesn't do any filtering (see the `paranoia_scope` at the top of the Story model).  This just causes any deletes of `v3` series to be paranoid, and deletes of `v4` series to be "real".

kookster: I guess that is what I mean - why not make them all set deleted_at, why fully delete v4?

kookster: (b/c I don't have a problem with empty v4 series showing up in v3, this isn't like pieces/stories where it will break things)

cavis: I guess I thought we were moving away from `paranoid` in v4, to having things actually deleted.  Maybe I mis-remembered that conversation.

kookster: I am still a pretty big fan of paranoid, but I can be convinced.
That seems like a larger debate however, so let's punt on it / log it.
This is consistent with pieces/stories - let's go with consistency for now and readdress closer to a wider release if this is seeming like a pain point still.

Issue #153, opened 2016-09-14
Update individual-account memberships


cavis:
v3 now gives users a membership into their default/individual account.

- [x] Simplify accounts/networks/stories relations in `user.rb`, which no longer need the "OR individual-account-id"
- [x] Update `user.rb` to also create a default/individual account + membership for new users (though I'm not sure we allow creating users via CMS just yet).

cavis:
Related: seeing many users without individual accounts in production, which causes a 500 given the current code (looks for a user's `individual_account.id`) - https://cms.prx.org/api/v1/users/239444

Also distressing - there are some users with multiple individual accounts (though much fewer).

```
mysql> select count(*) from users where id not in (select opener_id from accounts where type = 'IndividualAccount');
+----------+
| count(*) |
+----------+
|    20806 |
+----------+
1 row in set (0.34 sec)

mysql> select count(*) from (select u.id, count(*) as count from users u join accounts a on (u.id = a.opener_id) where a.type = 'IndividualAccount' group by u.id) t where count > 1;
+----------+
| count(*) |
+----------+
|      224 |
+----------+
1 row in set (0.85 sec)
```

@kookster is that the sort of thing we need to clean up?

kookster:
I have looked at this  - it is a bit smaller problem than that:
```
mysql> select count(a.id) as ac, a.opener_id from accounts a join users u on (u.id = a.opener_id) where a.type = 'IndividualAccount' and a.deleted_at is null  group by a.opener_id having ac > 1;
+----+-----------+
| ac | opener_id |
+----+-----------+
|  4 |     70199 | 
|  5 |     70236 | 
|  2 |     70339 | 
|  2 |     70473 | 
|  2 |     72162 | 
|  3 |     72163 | 
|  4 |     72751 | 
|  2 |     78370 | 
+----+-----------+
8 rows in set (4.09 sec)

```

kookster:
data is fixed, forever!
```

mysql> select count(a.id) as ac, a.opener_id from accounts a join users u on (u.id = a.opener_id) where a.type = 'IndividualAccount' and a.deleted_at is null  group by a.opener_id having ac > 1;
Empty set (3.46 sec)
```

PR #154, opened 2016-09-14
individual accounts are only through memberships

kookster:
fixes https://github.com/PRX/cms.prx.org/issues/153

### DIFF
--- a/app/models/membership.rb
+++ b/app/models/membership.rb
@@ -1,9 +1,16 @@
 # encoding: utf-8
 
 class Membership < BaseModel
-
   belongs_to :user, -> { with_deleted }, touch: true
   belongs_to :account, -> { with_deleted }, touch: true
 
   scope :approved, -> { where(approved: true) }
+
+  after_commit :update_user_default_account, on: [:destroy]
+
+  def update_user_default_account
+    if user.default_account == account
+      user.update_attributes!(account_id: user.individual_account.id)
+    end
+  end
 end
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -10,24 +10,29 @@ class User < BaseModel
   has_one :image, -> { where(parent_id: nil) }, class_name: 'UserImage'
 
   has_many :memberships
-  has_many :member_accounts, through: :memberships, source: :account
+  has_many :accounts, through: :memberships, source: :account
   has_many :producers
+
+  after_commit :create_individual_account, on: [:create]
 
   def individual_account
     accounts.where('type = \'IndividualAccount\'').first
   end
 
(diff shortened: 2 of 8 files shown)


app/models/membership.rb
```diff
@@ -1,9 +1,16 @@
...
   belongs_to :user, -> { with_deleted }, touch: true
   belongs_to :account, -> { with_deleted }, touch: true
 
   scope :approved, -> { where(approved: true) }
+
+  after_commit :update_user_default_account, on: [:destroy]
+
+  def update_user_default_account
+    if user.default_account == account
+      user.update_attributes!(account_id: user.individual_account.id)
```
cavis: So if I'm reading this correctly ... you're reassigning the user's default account, in the case that they lost membership into some other (non-individual) account.

app/models/user.rb
```diff
@@ -8,16 +8,32 @@ class User < BaseModel
...
+
+  def individual_account
+    accounts.where('type = \'IndividualAccount\'').first
+  end
+
+  def individual_account=(account)
+    prior_account = individual_account
+    memberships.create!(account_id: account.id, approved: true, role: 'admin')
+    if prior_account
+      memberships.where(account_id: prior_account.id).each { |m| memberships.destroy(m) }
```
cavis: In what case would a user have a prior individual account, that needed to be destroyed?

test/factories/user_factory.rb
```diff
@@ -1,18 +1,14 @@
...
     last_name 'Astley'
 
-    after(:build) do |user|
-      user.default_account = user.individual_account
-    end
-
-    after(:create) do |user|
-      create(:membership, user: user, account: user.individual_account)
+    after(:create) do |user, _evaluator|
+      user.individual_account = create(:individual_account)
```
cavis: won't the `after_commit` hook already take care of this?

app/models/user.rb
kookster: maybe, or maybe not - I could be updating that account into a group account or moving it to a different user. I don't want to assume it has to be destroyed.

app/models/membership.rb
kookster: yes. the logic is taken from V3. If you lose membership to the account that is currently your default, we update the default to your individual account rather than having it continue to point to an account you can't access, or to no account at all.

test/factories/user_factory.rb
kookster: sorta. The `after_commit` creates a basic, empty account. When using the factory to create the account, it gets additional data like stories in it which tests depend on.

cavis:
👍

PR #155, opened 2016-09-15
Audio version and file templates

### LINKED ISSUE #147: add audio file/version templates
Add templates to the API

kookster:
fixes #147 

- [x]  Models
- [x]  Controllers
- [x]  Representers
- [x]  Validations
- [x]  Add template to `AudioVersion`
- [x]  Code styles validate

(Next PR will take on audio version/file validations, but will further incorporate idea of different distributions with their own validations (e.g. b'cast has different audio and data requirements than a podcast))

### DIFF
--- a/app/models/audio_file_template.rb
+++ b/app/models/audio_file_template.rb
@@ -2,6 +2,7 @@
 
 class AudioFileTemplate < BaseModel
   belongs_to :audio_version_template
+
   acts_as_list scope: :audio_version_template
 
   before_validation :set_defaults, on: :create
--- a/app/models/audio_file_template.rb
+++ b/app/models/audio_file_template.rb
@@ -4,11 +4,19 @@ class AudioFileTemplate < BaseModel
   belongs_to :audio_version_template
   acts_as_list scope: :audio_version_template
 
+  before_validation :set_defaults, on: :create
+
   validates :length_minimum,
-    presence: true,
-    numericality: { less_than_or_equal_to: :length_maximum }
+            presence: true,
+            numericality: { less_than_or_equal_to: :length_maximum }
 
   validates :length_maximum,
-    presence: true,
-    numericality: { greater_than_or_equal_to: :length_minimum }
+            presence: true,
+            numericality: { greater_than_or_equal_to: :length_minimum }
+
+  def set_defaults
+    self.label ||= 'segment'
+    self.length_minimum ||= 0
+    self.length_maximum ||= 0
+  end
 end
--- a/app/models/audio_file_template.rb
+++ b/app/models/audio_file_template.rb
@@ -1,0 +1,5 @@
+# encoding: utf-8
+
+class AudioFileTemplate < BaseModel
+  belongs_to :audio_version_template
+end
--- a/app/models/audio_file_template.rb
+++ b/app/models/audio_file_template.rb
@@ -2,4 +2,13 @@
 
 class AudioFileTemplate < BaseModel
   belongs_to :audio_version_template
+  acts_as_list scope: :audio_version_template
+
+  validates :length_minimum,
+    presence: true,
+    numericality: { less_than_or_equal_to: :length_maximum }
+
+  validates :length_maximum,
+    presence: true,
+    numericality: { greater_than_or_equal_to: :length_minimum }
 end
--- a/app/models/audio_version.rb
+++ b/app/models/audio_version.rb
@@ -5,10 +5,7 @@ class AudioVersion < BaseModel
              class_name: 'Story',
              foreign_key: 'piece_id', touch: true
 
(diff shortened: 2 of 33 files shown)


app/models/audio_version_template.rb
```diff
@@ -0,0 +1,31 @@
...
+
+  validates :segment_count,
+            presence: true,
+            numericality: { only_integer: true, greater_than: 0 }
+
+  def sync_audio_file_templates
+    diff = segment_count - audio_file_templates.count
+    if diff > 0
+      diff.times { audio_file_templates.create! }
+    elsif diff < 0
```
kookster: hound is wrong, this throws an error

app/models/audio_version_template.rb
```diff
@@ -0,0 +1,31 @@
...
+            presence: true,
+            numericality: { greater_than_or_equal_to: :length_minimum }
+
+  validates :segment_count,
+            presence: true,
+            numericality: { only_integer: true, greater_than: 0 }
+
+  def sync_audio_file_templates
+    diff = segment_count - audio_file_templates.count
+    if diff > 0
```
kookster: hound is wrong, this throws an error

kookster:
What do you think @cavis? I think you should review, but I thought I could also review this with @sandikbarr as a way to start getting a handle on a fairly modern PRX rails app.

cavis:
I say the more people we have judging @kookster the better.

kookster:
@cavis you want to have a look now as well?

app/models/audio_file_template.rb
```diff
@@ -0,0 +1,22 @@
...
+  validates :length_minimum,
+            presence: true,
+            numericality: { less_than_or_equal_to: :length_maximum }
+
+  validates :length_maximum,
+            presence: true,
+            numericality: { greater_than_or_equal_to: :length_minimum }
+
+  def set_defaults
+    self.label ||= 'segment'
```
cavis: This is the default label for a new audio-file within this series?

app/models/audio_file_template.rb
```diff
@@ -0,0 +1,22 @@
...
+    self.length_minimum ||= 0
+    self.length_maximum ||= 0
```
cavis: Also - what is implied by min and max length of 0?  That it will enforce having 0 audio-files?

app/models/audio_file_template.rb
```diff
@@ -0,0 +1,22 @@
+# encoding: utf-8
+
+class AudioFileTemplate < BaseModel
```
cavis: Do audio file/version templates need a Policy as well?

app/models/audio_version.rb
```diff
@@ -1,26 +1,21 @@
 # encoding: utf-8
 
 class AudioVersion < BaseModel
+  belongs_to :story, -> { with_deleted },
+             class_name: 'Story',
+             foreign_key: 'piece_id', touch: true
+
+  belongs_to :template, -> { with_deleted },
```
cavis: It may also be convenient to return a link to the audio-version's template in the representer.

app/models/audio_file_template.rb
kookster: yeah, good point, also looks like some other models (AudioFile has a policy class, but not associated to the model...) are lacking a policy - let me do an audit and clean that up real quick.

app/models/audio_file_template.rb
kookster: Yeah, not deeply invested in this - the code in V3 is getting over the "not null" constraint by setting all labels to empty string. That drives me crazy, so I decided to give a bland value that could be overridden with something more colorful like "billboard" or "segment a" later on.

Do you have a better suggestion for a default?

app/models/audio_file_template.rb
kookster: min or max of 0 are ignored, treated as there being nothing set, so you can't have a min of 0.
this is consistent with v3 implementation.

app/models/audio_version.rb
kookster: yeah, I can add that.

app/models/audio_file_template.rb
cavis: Nah - just making sure I understood that.  For Publish, label will probably be a required field anyways.  So they'll have to create a non-default one.

app/models/audio_file_template.rb
kookster: This is turning out to be bigger than I wanted - I'm going to fix the template related policies, then do a 2nd pr with the clean up for the rest of the models

kookster:
@cavis I think that should satisfy everything.

cavis:
👍. :ship: :it:!

PR #158, opened 2016-09-20
Fix story controller multiple-filters

cavis:
The `noseries` filter was masking parent controller filters.  So I couldn't combine `?filters=noseries,v4`.


kookster:
makes sense.

PR #160, opened 2016-09-21
Add missing policy classes, tests

kookster:


kookster:
@cavis can you review? @sandikbarr pls look if you want/have time as well, but it is a very boring PR

Issue #156, opened 2016-09-19
Filtering series by text in name and description


kookster:

kookster:
@cavis I'm trying to decide if this is its own param, like `page`, `per` or `zoom`, or a filter param like `v4`, or `highlighted`.

My thought is probably a filter, where top level params affect the result set size and resolution, but the filter params affect which records ought to be included - is that how you see it as well?

cavis:
Yeah - continuing to add them to the `?filters` query param makes the responsibilities clear.  And at some point, I was thinking we could reveal the valid filters for a resource via the API links themselves.  By having the representer interrogate the controller, or something.

kookster:
yeah, could add them in some kind of response, OPTIONS maybe, or link attributes in the HAL resource?

PR #161, opened 2016-09-22
Filter by text for series and stories

kookster:
The ticket #156 was only for series, but went ahead and added the same for stories.

- [x] Scope added for `match_text(text)` to models
- [x] Controllers declare `text` filter param, use match_text scope for filter text to get resources

Use these by adding a `filters` param to the get request:
`https://cms.prx.org/api/v1/accounts/8/series?filters=text%3Dweirdos`
(where `%3D` decodes to `=`)

The series/stories will be filtered to match case-insensitively in the title, short description, or description against the text `weirdos`.

### DIFF
--- a/test/models/series_test.rb
+++ b/test/models/series_test.rb
@@ -79,6 +79,18 @@ describe Series do
       Series.where(id: [series.id, v3_series.id]).v4.must_include series
       Series.where(id: [series.id, v3_series.id]).v4.wont_include v3_series
     end
+
+    it 'searches text for title and description' do
+      series = create(:series,
+                      title: 'Some Weirdo',
+                      description: 'Unique thing',
+                      short_description: 'Lacking sense')
+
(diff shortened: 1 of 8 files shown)


test/models/series_test.rb
```diff
@@ -79,6 +79,18 @@
       Series.where(id: [series.id, v3_series.id]).v4.must_include series
       Series.where(id: [series.id, v3_series.id]).v4.wont_include v3_series
     end
+
+    it 'searches text for title and description' do
+      series = create(:series,
+                      title: 'Some Weirdo',
+                      description: 'Unique thing',
+                      short_description: 'Lacking sense')
```
sandikbarr: Maybe it's the Talking Heads blaring from the basement, but this makes perfect sense to me.

kookster: Yeah, I wish I could channel David Byrne, this may be as close as I ever get

Issue #145, opened 2016-07-15
add more generic implementation of sorting


kookster:
- [x] Parse list of columns and directions in new `sorts` request param
- [x] Specify allowed sorts by controller
- [x] Specify default sort for a controller
- [x] Add sorts to resource queries in controller
- [x] Add app level sorts and default to 
- [x] Add model specific sort values for series and story controllers

kookster:
this got merged!

PR #162, opened 2016-09-27
Add sorting to all resources

kookster:
implements #145

### DIFF
--- a/app/controllers/api/auth/stories_controller.rb
+++ b/app/controllers/api/auth/stories_controller.rb
@@ -1,14 +1,13 @@
 # encoding: utf-8
 
 class Api::Auth::StoriesController < Api::StoriesController
-
   include ApiAuthenticated
 
   api_versions :v1
 
   filter_resources_by :account_id, :series_id, :network_id
 
-  filter_params :noseries
+  filter_params :highlighted, :purchased, :v4, :noseries
 
   announce_actions :create, :update, :delete, :publish, :unpublish
 
@@ -20,10 +19,6 @@ class Api::Auth::StoriesController < Api::StoriesController
     user_not_authorized unless current_user.networks.exists?(params[:network_id])
   end
 
-  def sorted(res)
-    res.order('updated_at desc')
-  end
-
   # ALL stories - not just published and visible
   def scoped(relation)
     relation
@@ -31,7 +26,7 @@ class Api::Auth::StoriesController < Api::StoriesController
 
   def filtered(resources)
     resources = resources.unseries if filters.noseries?
-    super
+    super(resources)
   end
 
   def resources_base
@@ -51,5 +46,4 @@ class Api::Auth::StoriesController < Api::StoriesController
       story.account_id ||= current_user.approved_accounts.first.try(:id)
     end
   end
-
 end
--- a/app/controllers/api/auth/stories_controller.rb
+++ b/app/controllers/api/auth/stories_controller.rb
@@ -9,6 +9,10 @@ class Api::Auth::StoriesController < Api::StoriesController
 
   filter_params :highlighted, :purchased, :v4, :text, :noseries
 
+  sort_params default: { updated_at: :desc },
+              allowed: [:id, :created_at, :updated_at, :published_at, :title,
+                        :episode_number, :position]
+
   announce_actions :create, :update, :delete, :publish, :unpublish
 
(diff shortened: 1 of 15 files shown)


kookster:
finally finished this one up, yeesh @cavis 
I'll schedule review time with @sandikbarr

app/controllers/api/auth/stories_controller.rb
```diff
@@ -1,14 +1,13 @@
...
 class Api::Auth::StoriesController < Api::StoriesController
-
   include ApiAuthenticated
 
   api_versions :v1
 
   filter_resources_by :account_id, :series_id, :network_id
 
-  filter_params :noseries
+  filter_params :highlighted, :purchased, :v4, :text, :noseries
```
cavis: Also, it's slightly strange to define the `filter_params` here, when it's the filtered `super(resources)` call that actually applies them.

app/controllers/api/stories_controller.rb
```diff
@@ -74,11 +77,8 @@ def scoped(relation)
   end
 
   def sorted(res)
-    if filters.purchased?
-      res.purchased.order('purchase_count DESC')
-    else
-      res.order('published_at desc')
-    end
+    res = res.purchased.order('purchase_count DESC') if filters.purchased?
```
cavis: Just to double check ... this `order` will get applied before the default sort, correct?

test/models/tagging_test.rb
```diff
@@ -15,6 +15,10 @@
     tagging.must_respond_to(:taggable)
   end
 
+  it 'has an owner' do
+    tagging.owner.wont_be_nil
```
cavis: Off-topic?

kookster: yes it absolutely is, I just noticed some missing unit tests in the code coverage and put that in as a bonus.

app/controllers/api/auth/stories_controller.rb
cavis: Yeah, this is probably the more explicit option, then.  So be it!

app/controllers/api/auth/stories_controller.rb
```diff
@@ -20,18 +19,14 @@ def check_user_network
     user_not_authorized unless current_user.networks.exists?(params[:network_id])
   end
 
-  def sorted(res)
-    res.order('updated_at desc')
```
cavis: I think that may be a good default for the API.  When viewing all stories ... you get them in order of published date.  When viewing only your stories (including unpublished), you get them in order of updated_at.

app/controllers/api/auth/stories_controller.rb
```diff
@@ -1,14 +1,17 @@
...
   include ApiAuthenticated
 
   api_versions :v1
 
   filter_resources_by :account_id, :series_id, :network_id
 
-  filter_params :noseries
+  filter_params :highlighted, :purchased, :v4, :text, :noseries
+
+  sort_params default: { updated_at: :desc },
```
kookster: default is now updated_at @cavis

app/controllers/api/auth/stories_controller.rb
kookster: fixed above.

app/controllers/api/auth/stories_controller.rb
cavis: Nice.  And it still inherits `allowed` from the parent?

kookster: no, when I call `sort_params`, I have to specify both, so line 13 has the allowed sort params.
(I didn't implement partial overrides.)

cavis: Gotcha.

cavis:
Looks good to me.  Unassigning myself!

app/controllers/concerns/api_sorting.rb
```diff
@@ -0,0 +1,55 @@
...
+
+  private
+
+  def parse_sorts_param
+    sorts_array = []
+    allowed_sorts = self.class.allowed_sort_names || self.class.superclass.allowed_sort_names
+
+    # parse sort param for name of the column and direction
+    # default is descending, because I say so, and we have a bias towards the new
+    (params[:sorts] || '').split(',').each do |str|
```
sandikbarr: add a comment example of this: /?sorts=published_at:desc,updated_at:desc

PR #164, opened 2016-09-27
Fix representers to have consistent contents for embedded stories

kookster:
- [x] Implement standard method for what are public stories, using story scopes
- [x] Change default story relations to include all stories, and add public filters as needed in representers and controllers, matching in both. This essentially renames `all_stories` to `stories`, and adds a `public_stories` method to indicate when only stories visible to non-authenticated users will be returned
- [x] Add standard url template suffix for supported query params, like page, per, zoom, sorts, filters

### DIFF
--- a/app/models/account.rb
+++ b/app/models/account.rb
@@ -9,8 +9,7 @@ class Account < BaseModel
   has_one :image, -> { where(parent_id: nil) }, class_name: 'AccountImage'
   has_one :portfolio
 
-  has_many :stories, -> { where('published_at is not null and network_only_at is null').order(published_at: :desc) }
-  has_many :all_stories, foreign_key: 'account_id', class_name: 'Story'
+  has_many :stories, -> { order(published_at: :desc) }
   has_many :series
   has_many :memberships
   has_many :websites, as: :browsable
--- a/app/models/account.rb
+++ b/app/models/account.rb
@@ -1,6 +1,8 @@
 # encoding: utf-8
 
 class Account < BaseModel
+  include Storied
+
   acts_as_paranoid
 
   belongs_to :opener, -> { with_deleted }, class_name: 'User', foreign_key: 'opener_id'
--- a/app/models/base_model.rb
+++ b/app/models/base_model.rb
@@ -14,8 +14,4 @@ class BaseModel < ActiveRecord::Base
     raw_write_attribute(:filename, name)
     save!
   end
-
-  def public_stories(arel = nil)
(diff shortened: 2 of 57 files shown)


app/representers/api/base_representer.rb
```diff
@@ -17,4 +17,8 @@ def self.alternate_host
   def self.profile_host
     ENV['META_HOST'] || 'meta.prx.org'
   end
+
+  def index_url_params
```
cavis: Nice!

Would it also be worth adding a `show_url_params`?  Or maybe we should just wait until there's more than `?zoom` in there.  Stop pre-optimizing, Ryan!

kookster: might as well add it, while I'm in here making things sweet.

app/models/base_model.rb
```diff
@@ -14,4 +14,8 @@ def update_file!(name)
     raw_write_attribute(:filename, name)
     save!
   end
+
+  def public_stories(arel = nil)
```
kookster: we need the method in series, network, account, and playlist (I think), which seemed like enough to justify putting it in there, but you know, maybe I should make a simple concern instead...

cavis: Ah, gotcha - was forgetting that we often access story-counts from the other side of the relation.  (`series.public_stories.count` ... instead of `Story.public_stories.where(series: series).count`).  Yeah, I think having it here is fine.

app/representers/api/base_representer.rb
kookster: added!

app/models/base_model.rb
kookster: added a `Storied` concern for models with `has_many :stories`

app/models/account.rb
```diff
@@ -9,8 +9,7 @@ class Account < BaseModel
   has_one :image, -> { where(parent_id: nil) }, class_name: 'AccountImage'
   has_one :portfolio
 
-  has_many :stories, -> { where('published_at is not null and network_only_at is null').order(published_at: :desc) }
-  has_many :all_stories, foreign_key: 'account_id', class_name: 'Story'
+  has_many :stories, -> { order(published_at: :desc) }
```
kookster: I don't think we are, but the public account representer correctly uses public_stories which will match the stories_controller using the same with a filter by account_id

kookster:
@cavis I think this is good to go now

cavis:
:+1:

PR #167, opened 2016-09-29
Fix image routes

kookster:
- [x] Get rid of direct access to image resources (use nested only)
- [x] Don't return the series image if a story has none
- [x] Don't use polymorphic_url when it isn't needed
- [x] Don't add a profile attribute if the link is a consistent type
- [x] Add helper method for image representers to resolve url to image resource

### DIFF
--- a/app/controllers/api/series_images_controller.rb
+++ b/app/controllers/api/series_images_controller.rb
@@ -1,10 +1,17 @@
 # encoding: utf-8
 
 class Api::SeriesImagesController < Api::BaseController
-
   api_versions :v1
 
   filter_resources_by :series_id
 
   represent_with Api::ImageRepresenter
+
+  def resource
+    @series_image ||= series.try(:image) || super
+  end
+
+  def series
+    @series ||= Series.find(params[:series_id]) if params[:series_id]
+  end
 end
--- a/app/representers/api/msg/image_representer.rb
+++ b/app/representers/api/msg/image_representer.rb
@@ -4,6 +4,20 @@
 # Representer for Announce messaging
 #
 class Api::Msg::ImageRepresenter < Api::BaseRepresenter
+  def self_url(represented)
+    if represented.is_a?(StoryImage)
+      api_story_story_image_path(represented.story, represented)
+    elsif represented.is_a?(SeriesImage)
+      api_series_series_image_path(represented.series)
+    elsif represented.is_a?(AccountImage)
+      api_account_account_image_path(represented.account)
+    elsif represented.is_a?(UserImage)
+      api_user_user_image_path(represented.user)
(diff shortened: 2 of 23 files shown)


app/controllers/api/series_images_controller.rb
```diff
@@ -1,10 +1,17 @@
...
 class Api::SeriesImagesController < Api::BaseController
-
   api_versions :v1
 
   filter_resources_by :series_id
 
   represent_with Api::ImageRepresenter
+
+  def resource
+    @series_image ||= series.try(:image) || super
```
cavis: Did you confirm what announce will/won't do, when CRUD-ing these series/account/user images?

Now that I look, only StoryImages are announced at the moment.  Will need to add the others - maybe a separate issue/PR?

app/representers/api/msg/image_representer.rb
```diff
@@ -4,6 +4,12 @@
 # Representer for Announce messaging
 #
 class Api::Msg::ImageRepresenter < Api::BaseRepresenter
+  include NestedImage
+
+  def self_url(represented)
```
cavis: The announce-representer doesn't actually need this.  But I guess it can't hurt for it to know the CMS-url for the image as well.

test/representers/api/story_representer_test.rb
```diff
@@ -64,15 +64,6 @@
       json['_links']['profile']['href'].must_equal 'http://meta.prx.org/model/story'
     end
 
-    it 'has a profile for the default image' do
```
cavis: Is it no longer "usual" to include a profile link on the image links?  (Looks like you've standardized them here to include `href` and `title`, but not `profile`).

cavis:
👍

app/controllers/api/series_images_controller.rb
kookster: yes, different PR I think.

app/representers/api/msg/image_representer.rb
kookster: The hal base representer will always try to add a self link, so this makes it not fail to generate.
Before it was adding one w/o error using the direct routes to the images.

test/representers/api/story_representer_test.rb
kookster: profile is aways provided on the resource itself, but the link relation is supposed to be enough info about the linked resource. When we have had links to resources where there are many possible resources, and so the profile is not predictable or constant, we have specified a profile attribute with the other link attributes.

Now that the images will always be consistent about what profile they are for a given parent resource's link relation, no need to have profile in there - you'll see usually we don't have profiles for the vast majority of links (except self).

PR #168, opened 2016-09-29
Update user policy, improve test coverage

kookster:
# !!! Review and merge https://github.com/PRX/cms.prx.org/pull/167 first please. !!!

- [x] prevent create on user (this needs to be done with id.prx.org, just block it for now)
- [x] add test class for schedule model
- [x] improve policy testing
- [x] make fixerable only support aws (really only expect to need s3)
- [x] add and move tests for model concerns


cavis:
👍

PR #169, opened 2016-09-29
Add in announce for image controllers

kookster:


cavis:
98.4% coverage, huh?  Couldn't get that to 98.5?

kookster:
well, I haven't added tests for these yet, so, lookout, here comes 98.6, which means our app is not running a fever.

kookster:
@cavis looks like I have some hounding to fix...

kookster:
@cavis I think this is good now, hopefully I didn't forget anything else...

app/controllers/concerns/child_resource.rb
```diff
@@ -0,0 +1,69 @@
...
+  end
+
+  def create_child_resource
+    create_resource.tap do |res|
+      original = child_resource
+      opts = create_options
+      consume! res, opts
+      hal_authorize res
+      parent_resource.public_send("#{child_resource_name}=", res)
+      try(:after_original_destroyed, original) if original.destroyed?
```
cavis: Could `original` possibly be `nil` here?  (If creating where one didn't exist before).

app/controllers/api/user_images_controller.rb
```diff
@@ -1,18 +1,19 @@
...
   filter_resources_by :user_id
 
+  announce_actions decorator: Api::Msg::ImageRepresenter, subject: :image
+
   represent_with Api::ImageRepresenter
 
-  def resource
-    @user_image ||= user.try(:image) || super
-  end
+  child_resource parent: 'user', child: 'image'
```
cavis: 👍  some fancy coding, to make this happen!

app/representers/api/account_representer.rb
```diff
@@ -18,11 +18,18 @@ class Api::AccountRepresenter < Api::BaseRepresenter
...
+      title: represented.image.filename
+    } if represented.id && represented.image
   end
   embed :image, class: Image, decorator: Api::ImageRepresenter
 
+  link 'create-image' do
+    {
+      href: api_account_account_image_path(represented),
+      title: 'Create an image'
+    } if represented.id && !represented.image
```
cavis: What's the case for not always including this link?  Just to cut down on payload?  (I myself may have made that argument).

app/workers/audio_callback_worker.rb
```diff
@@ -14,7 +14,7 @@ def perform(_sqs_msg, job)
     audio.frequency = (job['frequency'] || 0) / 1000.0
 
     # decode content type and mpeg layer from basic "format" string
-    mime_types = MIME::Types.type_for(job['format']).map(&:to_s)
+    mime_types = MIME::Types.type_for(job['format'] || '').map(&:to_s)
```
cavis: For some reason, I was thinking that `type_for` would handle the `nil`.  But this is better.

test/controllers/concerns/child_resource_test.rb
```diff
@@ -0,0 +1,130 @@
...
+  Rails.application.routes.draw do
+    namespace :api do
+      resources :moms do
+        resource :son
+      end
+    end
+  end
+end
+
+describe Api::SonsController do
```
cavis: Wow, lots of test setup to get to this point!

app/representers/api/account_representer.rb
kookster: Once you have an `image` link, you don't need the `create-link`, you can PUT to that and replace it entirely.

app/workers/audio_callback_worker.rb
kookster: it used to, when I upgraded the gem, it no longer does, it will return `nil` instead of `[]`.

test/controllers/concerns/child_resource_test.rb
kookster: I know, srsly, I do have some test models I use for this sort of thing, but they weren't quite right:
https://github.com/PRX/cms.prx.org/blob/master/test/test_models.rb

Maybe I should add the test setup to `test_models` instead of in this test?

cavis: I think this is fine for now.. until a 2nd spec starts doing all this setup.

app/controllers/concerns/child_resource.rb
kookster: fixed.

kookster:
@cavis I made one small edit post review, good enough?

cavis:
👍

PR #182, opened 2016-10-18
Series templates

kookster:


app/representers/api/series_representer.rb
```diff
@@ -47,4 +47,14 @@ class Api::SeriesRepresenter < Api::BaseRepresenter
     }
   end
   embed :account, class: Account, decorator: Api::Min::AccountRepresenter
+
+  link :audio_version_templates do
+    {
+      href: api_series_audio_version_templates_path(represented)
```
cavis: Can I argue to also put a `count` on this, to save me a request on the client side?  Or am I just being lazy?

cavis: Mind changed ... caring about series-templates is a bit of an edge case.  An extra request doesn't really hurt anything.

PR #188, opened 2016-11-15
Distributions

kookster:
- [x] relies on db changes to prx.org PRX/prx.org/pull/287


lib/prx_access.rb
```diff
@@ -0,0 +1,82 @@
...
+    {
+      'Content-Type' => 'application/json',
+      'Accept' =>  'application/json'
+    }
+  end
+
+  def api(options = {})
+    opts = { root: cms_root, headers: default_headers }.merge(options)
+    if account = opts.delete(:account)
+      token = get_account_token(account)
```
cavis: I'm guessing this doesn't get used often enough to be worth caching across requests.  Right?

config/initializers/config_env.rb
```diff
@@ -0,0 +1,8 @@
+# default env vars that may not be set
+ENV['ID_ROOT'] ||= 'https://id.prx.org/'
+ENV['CMS_ROOT'] ||= 'https://cms.prx.org/api/v1/'
+ENV['PRX_ROOT'] ||= 'https://beta.prx.org/stories/'
+ENV['CRIER_ROOT'] ||= 'https://crier.prx.org/api/v1'
```
cavis: No `FEEDER_ROOT` here?

app/controllers/api/distributions_controller.rb
```diff
@@ -0,0 +1,17 @@
+# encoding: utf-8
+
+class Api::DistributionsController < Api::BaseController
+  api_versions :v1
+
+  filter_resources_by :series_id
+
+  announce_actions resource: :owner_resource
```
cavis: Still not sure I follow.  So if this is a `PodcastDistribution`, this controller will announce changes to it with the `Series`?

app/controllers/concerns/polymorphic_resource.rb
```diff
@@ -0,0 +1,28 @@
+# encoding: utf-8
+
+require 'active_support/concern'
+
+module PolymorphicResource
+  extend ActiveSupport::Concern
+
+  def polymorphic_filtered(polymorphic, arel)
```
cavis: Fancy!

app/models/distribution.rb
```diff
@@ -0,0 +1,20 @@
+# encoding: utf-8
+require 'hash_serializer'
+
+class Distribution < BaseModel
+  belongs_to :distributable, polymorphic: true, touch: true
```
cavis: Polymorphics still cascade-delete, right?  #toolazytolookitup

app/models/series.rb
```diff
@@ -18,6 +18,7 @@ class Series < BaseModel
   has_many :stories, -> { order('episode_number DESC, position DESC, published_at DESC') }
   has_many :schedules
   has_many :audio_version_templates
+  has_many :distributions, as: :distributable
```
cavis: Why rename?  I thought the Series _was_ "distributable", and the PodcastDistribution is the "distribution of the distributable".

app/controllers/api/distributions_controller.rb
cavis: I must have something wrong about what `resource:` does, as the `DistributionRepresenter` looks like it's using Distribution fields.

app/representers/api/distribution_representer.rb
```diff
@@ -0,0 +1,20 @@
+# encoding: utf-8
+
+class Api::DistributionRepresenter < Api::BaseRepresenter
+  property :id, writeable: false
+  property :type
+  property :guid
+  property :url
+  hash :properties
```
cavis: A nested-object inside a HAL resource, huh?  I don't see anything in the spec prohibiting it, but there are some comments in the `hal-discuss` google group frowning on it.  Opinions?  Do we do this anywhere else in CMS?

config/initializers/config_env.rb
cavis: Should probably also add to the `env-example`.

cavis: Also ... why are these `_ROOT` here, but `_HOST` in the `env-example`?

app/controllers/api/audio_files_controller.rb
```diff
@@ -1,12 +1,11 @@
 # encoding: utf-8
 
 class Api::AudioFilesController < Api::BaseController
-
   api_versions :v1
 
   filter_resources_by :audio_version_id, :account_id
 
-  announce_actions decorator: Api::Msg::AudioFileRepresenter, subject: :audio
+  announce_actions subject: :audio
```
kookster: Yeah, made it a bit smarter for our purposes, you can still override if you need to.

app/controllers/api/distributions_controller.rb
kookster: Let me preface this by saying, this may be a bad idea.
When a distribution gets updated/created/deleted, I want an announcement to go out that the owner of that distribution has been changed (i.e. the series), so anything listening to changes for the series will realize there are changes to the series' distributions (oh, this is a podcast now!)

lib/prx_access.rb
kookster: It is a concern that this could be a bit slow and slow down http responses, but it is not frequently a synchronous call.

What do you want think we could cache here? Once the `api` method returns the hypermedia api client, it can be used for multiple requests with the same access token for that controller request, are you thinking to hold on to the access token or root doc or something between controller requests?

app/models/distribution.rb
kookster: I am also lazy but I think so if they are configured for it.

app/models/series.rb
kookster: I'm not renaming it, I'm using the polymorphic relationship to the distribution.
I am perhaps over engineering, but I expect some forms of distribution may be at an account level instead of the series, so not only series can be `distributable`, i.e. own distributions.

app/representers/api/distribution_representer.rb
kookster: If it isn't an addressable, stand alone resource, then I think this is an ok way to handle it.
If it was able to be updated distinctly from the distribution, as its own resource, then no.
Do you have any other ideas?

PR #189, opened 2016-11-17
Public asset fixes

cavis:
Addresses a couple of issues with public assets for Publish...

- CORS headers not being set on `/pub/` paths, preventing Publish from following the redirect for non-native (aurora) mp3 playback
- HEAD requests for `/pub/` S3 assets were not being signed correctly
- No OPTIONS set for `/pub/`

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -85,9 +85,10 @@ PRX::Application.routes.draw do
   match '/', via: [:get], to: redirect("/api/v1")
 
   public_asset = 'pub/:token/:expires/:use/:class/:id/:version/:name.:extension'
-  public_name = {name: /[^\/]+/}
+  public_name = { name: /[^\/]+/ }
   match public_asset, to: 'public_assets#show_head', via: :head, constraints: public_name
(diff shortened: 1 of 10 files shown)


config/routes.rb
```diff
@@ -85,5 +85,6 @@
   match '/', via: [:get], to: redirect("/api/v1")
 
   get 'pub/:token/:expires/:use/:class/:id/:version/:name.:extension' => 'public_assets#show', as: :public_asset, constraints: {name: /[^\/]+/}
+  match 'pub/*any', via: [:options], to: 'public_assets#options'
```
kookster: Why not also make a method for head, as you did for options? seems inconsistent, idk.

kookster: Why not also make a method for head, as you did for options? seems inconsistent, idk.

cavis: I could be convinced of that.  Would have to duplicate (or more likely abstract) some of that shared HEAD/GET logic in the current `show` method.

PR #190, opened 2016-11-30
Configurable ID_HOST

cavis:
Makes ID_HOST explicitly configurable, to help with local testing of Publish + ID + CMS.


cavis:
@kookster minimal changes here - just makes developing locally easier.  And will allow setting up `id-staging.prx.tech` with just a few ENV changes to cms-staging.

PR #196, opened 2016-12-08
Fix up the distributions

kookster:
- [x] Fix the API calls to be able to create series podcast distribution


app/controllers/api/distributions_controller.rb
```diff
@@ -14,4 +17,26 @@ def owner_resource
...
     polymorphic_filtered('distributable', arel)
   end
+
+  def create
+    create_resource.tap do |res|
+      consume! res, create_options
+      res = res.becomes(res.type.safe_constantize) if res.type
+      hal_authorize res
+      res.save!
+      distribute(res)
```
cavis: I see what you're up to here... sort of a manual after-save-but-before-response hook.

The repeated HAL boilerplate above this does catch my attention.  Seems to be shared here and in the story-distributions controller.  Maybe relocate to a shared concern, along with the `decorator_for_model` stuff?

app/models/distributions/podcast_distribution.rb
```diff
@@ -1,30 +1,31 @@
...
-module Distributions
-  class PodcastDistribution < ::Distribution
-    include PRXAccess
-    include Rails.application.routes.url_helpers
+class Distributions::PodcastDistribution < Distribution
+  include PRXAccess
+  include Rails.application.routes.url_helpers
 
-    after_commit :add_podcast_to_feeder, on: [:create]
+  def distribute
```
cavis: Sticking with the activerecord norms ... would this be `distribute!` instead of `distribute`?  Or is that not hip anymore?

app/models/story_distributions/episode_distribution.rb
```diff
@@ -1,15 +1,18 @@
...
   include PRXAccess
   include Rails.application.routes.url_helpers
 
-  after_commit :add_episode_to_feeder, on: [:create]
+  def distribute
+    super
+    add_episode_to_feeder
+  end
 
   def add_episode_to_feeder
```
cavis: Off-topic: how are podcasts/episodes removed from feeder (after deletion from CMS)?

app/representers/api/distribution_representer.rb
```diff
@@ -2,9 +2,9 @@
 
 class Api::DistributionRepresenter < Api::BaseRepresenter
   property :id, writeable: false
-  property :type
   property :guid
   property :url
+  property :kind
```
cavis: 👍 for non-overloaded (or at least less-overloaded) nouns!

db/schema.rb
```diff
@@ -454,16 +454,17 @@
   add_index "series_images", ["series_id"], name: "series_images_series_id_fk", using: :btree
 
   create_table "story_distributions", force: :cascade do |t|
+    t.string   "type",            limit: 255
     t.integer  "distribution_id", limit: 4
-    t.integer  "story_id",        limit: 4
+    t.integer  "piece_id",        limit: 4
```
cavis: But no more required schema changes to the staging DB right?  (I thought this one was already in there when I ran them).

app/controllers/api/distributions_controller.rb
kookster: yeah, wasn't sure this was worth refactoring yet, since I don't see other cases of this besides distributions.

app/models/story_distributions/episode_distribution.rb
kookster: when a story is deleted, it sends an `announce` notification which should remove from feeder.
the story distribution in feeder is not distinctly deleted, but since the parent story is deleted, it is effectively deleted as well.

app/representers/api/distribution_representer.rb
kookster: how very ...nice? of you to say

app/models/story_distributions/episode_distribution.rb
kookster: no, lies, nvm, still thinking.

app/controllers/api/distributions_controller.rb
kookster: no, lies, nvm, still thinking.

app/models/distributions/podcast_distribution.rb
cavis: True, this is all within an `update_attribute`, so not entirely obvious what this should do.  Non bang version seems like the right call for now, then.

kookster:
@cavis I think this is better now, pls have a 2nd (3rd, 4th?) look

test/controllers/concerns/abstract_resource_test.rb
```diff
@@ -0,0 +1,108 @@
...
+  def id
+    1
+  end
+
+  def initialize
+    self.woof = 'WOOF'
+  end
+end
+
+class Api::MammalsController < ActionController::Base
```
cavis: I read that as the API "(yo') mamma is" Controller.

app/controllers/concerns/abstract_resource.rb
```diff
@@ -0,0 +1,51 @@
+# encoding: utf-8
+
+require 'active_support/concern'
+
+module AbstractResource
```
cavis: It's all so encapsulated... I can hardly stand it!

PR #202, opened 2016-12-14
Fix Feeder API calls

kookster:
- [x] Camelize attributes before submitting
- [x] use .org not .tech for feeder
- [x] get rid of `embeds`, except where it will break things


kookster:
@cavis I am making hound happy, stay tuned

Issue #203, opened 2016-12-14
Distributions announcing to wrong SNS topic


cavis:
Seen in the cms-staging logs, on POSTing a new podcast distribution: 

```
Non-existent SNS topic: staging_announce_distribution_create
```

kookster:
fixed, bug where the `resource` option was filtered out.

PR #204, opened 2016-12-15
Fix publish podcast

kookster:
- [x] Make sure announce message publish for the owner resource (e.g. the series) not the distribution #203 
- [x] Send `published_at` and series specific data for initial create of a podcast in feeder
- [x] Additional test for deleting series image, since there was a question about cms handling

### DIFF
--- a/app/controllers/concerns/announce_actions.rb
+++ b/app/controllers/concerns/announce_actions.rb
@@ -40,7 +40,7 @@ module AnnounceActions
     end
 
     def new_announce_filter(action, options)
-      filter_options = options.slice(:action, :decorator, :subject)
+      filter_options = options.slice(:action, :decorator, :subject, :resource)
       filter_options[:action] ||= 'delete' if action.to_s == 'destroy'
       AnnounceActions::AnnounceFilter.new(action, filter_options)
     end
--- a/app/models/distributions/podcast_distribution.rb
+++ b/app/models/distributions/podcast_distribution.rb
@@ -23,9 +23,22 @@ class Distributions::PodcastDistribution < Distribution
   end
 
   def podcast_attributes
-    {
+    attrs = {
       prx_uri: polymorphic_url(['api', owner], only_path: true),
-      prx_account_uri: api_account_path(account)
+      prx_account_uri: api_account_path(account),
+      published_at: Time.now
     }
+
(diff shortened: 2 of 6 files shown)


app/controllers/concerns/announce_actions.rb
```diff
@@ -40,7 +40,7 @@ def add_announce_filter(action, options)
     end
 
     def new_announce_filter(action, options)
-      filter_options = options.slice(:action, :decorator, :subject)
+      filter_options = options.slice(:action, :decorator, :subject, :resource)
```
kookster: This is what fixes the announce issue, the new attribute was never getting to the annouce filter

app/models/distributions/podcast_distribution.rb
```diff
@@ -23,9 +23,22 @@ def get_podcast
   end
 
   def podcast_attributes
-    {
+    attrs = {
       prx_uri: polymorphic_url(['api', owner], only_path: true),
-      prx_account_uri: api_account_path(account)
+      prx_account_uri: api_account_path(account),
+      published_at: Time.now
```
kookster: I don't know a better value to set this to, so here goes - just needs to have a value.

test/controllers/concerns/announce_actions_test.rb
```diff
@@ -53,5 +39,42 @@
       last_message['subject'].to_s.must_equal 'test_object'
       last_message['action'].to_s.must_equal 'delete'
     end
+
+    it 'will call announce on a different resource' do
```
kookster: This is the new test, the rest is re-org

kookster:
@cavis added some comments to clarify what the heck I am doing here

app/models/distributions/podcast_distribution.rb
```diff
@@ -23,9 +23,22 @@ def get_podcast
...
+
+    if owner.is_a?(Series)
+      attrs[:title] = owner.title
+      attrs[:subtitle] = owner.short_description
+      attrs[:description] = owner.description
+      attrs[:summary] = owner.description
+      if owner.image
+        attrs[:itunes_image] = { url: owner.image.public_url(version: 'original') }
+      end
+    end
```
cavis: Agreed.

PR #205, opened 2016-12-15
Camelize resource attribute names in json requests

kookster:
- [x] monkey patch hyperresource to use outgoing_body_filter


kookster:
That test is still failing sometimes, let me see if I can fix it.

kookster:
@cavis I believe hound is as happy as it will be, since I have code copied from the `hyperresource` gem I don't want to mess with. It will get better if `hyperresource` gets fixed...

kookster:
@cavis I think that test will keep passing, but I've been wrong before!

cavis:
Woooo!  First CMS PR to ever pass all the checks!

PR #206, opened 2016-12-16
Use markdown for V4 stories and series, but continue to return html for description

kookster:
- [x] Add description_md attribute with md value of description
- [x] Render v4 descriptions from md -> html for description field
- [x] Render v3 descriptions from html ->  for description_md field
- [x] Allow updating the desc also as description_md

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -47,6 +47,8 @@ gem 'roar'
 gem 'roar-rails'
 gem 'oj'
 gem 'oj_mimic_json'
+gem 'redcarpet'
+gem 'reverse_markdown'
 
 ## Messaging
 gem 'shoryuken'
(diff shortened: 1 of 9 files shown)


kookster:
@sandikbarr added you on here as well so you can see what madness is happening with cms and markdown

kookster:
@sandikbarr @cavis this is the one that breaks how markdown is retrieved from the cms api - are we ready to do this and break publish?

sandikbarr:
@kookster @cavis I'm ready. I can open a PR for switching this field this afternoon.

sandikbarr:
@cavis @kookster Opened a Publish PR for this change https://github.com/PRX/publish.prx.org/pull/158. Not sure if I should have added another field for description_md. I just pointed it to the existing description member variable since we are not using the html formatted description anywhere. I opted for making the least amount of changes, but I can certainly add description_md as an additional variable and keep the html in the description field on the story and series model instead.

PR #208, opened 2016-12-17
Multiple series images

kookster:
- [x] Update models to add default image and has_many for series images
- [x] Update routes for series images to be resource(s)
- [x] Update representers
- [x] Add "use" or "purpose" attribute to model and filter for controller
- [x] Determine default image using purpose as well
- [x] Send the correct images to feeder based on `purpose`
- [x] Define common image purposes
- [x] Add filter by purpose to image controllers
- [x] Set both profile and thumbnail images on podcast

### DIFF
--- a/app/controllers/api/series_images_controller.rb
+++ b/app/controllers/api/series_images_controller.rb
@@ -1,8 +1,6 @@
 # encoding: utf-8
 
 class Api::SeriesImagesController < Api::BaseController
-  include Announce::Publisher
-
   api_versions :v1
 
   filter_resources_by :series_id
@@ -10,10 +8,4 @@ class Api::SeriesImagesController < Api::BaseController
   announce_actions decorator: Api::Msg::ImageRepresenter, subject: :image
 
   represent_with Api::ImageRepresenter
-
-  child_resource parent: 'series', child: 'image'
-
-  def after_original_destroyed(original)
-    announce('image', 'destroy', Api::Msg::ImageRepresenter.new(original).to_json)
-  end
 end
--- a/app/models/image.rb
+++ b/app/models/image.rb
@@ -12,6 +12,18 @@ class Image < BaseModel
   FAILED = 'failed'.freeze
   COMPLETE = 'complete'.freeze
 
+  PROFILE = 'profile'.freeze
+  THUMBNAIL = 'thumbnail'.freeze
+  PURPOSES = [PROFILE, THUMBNAIL]
+
+  def self.profile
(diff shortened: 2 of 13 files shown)


kookster:
I used `purpose` instead of `use` because `use` is a SQL reserved word and caused some headaches.

app/controllers/api/series_images_controller.rb
```diff
@@ -1,19 +1,11 @@
 # encoding: utf-8
 
 class Api::SeriesImagesController < Api::BaseController
-  include Announce::Publisher
```
cavis: Really simplifies this logic.

app/models/image.rb
```diff
@@ -12,6 +12,18 @@ class Image < BaseModel
   FAILED = 'failed'.freeze
   COMPLETE = 'complete'.freeze
 
+  PROFILE = 'profile'.freeze
+  THUMBNAIL = 'thumbnail'.freeze
+  PURPOSES = [PROFILE, THUMBNAIL].freeze
+
+  def self.profile
```
cavis: So you're adding the `purpose` field to every image type, then?

Also, is profile-vs-thumbnail a generic enough relation to warrant having them in every image sub-type?  (Mostly playing devil's advocate here)

app/representers/api/series_representer.rb
```diff
@@ -26,18 +26,19 @@ class Api::SeriesRepresenter < Api::BaseRepresenter
...
+  link :images do
     {
-      href: api_series_series_image_path(represented),
-      title: 'Create an image'
-    } if represented.id && !represented.image
+      href: api_series_series_images_path(represented),
+      count: represented.images.count
+    } if represented.id
   end
+  embed :images, paged: true, item_class: SeriesImage, item_decorator: Api::ImageRepresenter
```
cavis: Is it necessary to embed all series-images (usually 2) in this response?  I think the pattern for Story is to embed the default (the 1st by position), and just link/count the others.

app/models/image.rb
kookster: yes, I think so, I thought about it.

app/representers/api/series_representer.rb
kookster: Yeah, I updated this to embed them all b/c it is as common to the podcast as the audio files are to a story, i.e. I am going to end up in feeder having to retrieve them both every time there is an update. They are not all in the min representer, so when included in a story or a list they don't appear.

PR #210, opened 2016-12-20
Prevent podcast distributions with no podcast

### LINKED ISSUE #209: Do not create podcast distributions without a feeder podcast
Rollback  distro create if feeder podcast create fails

kookster:
fixes #209 
- [x] Raise errors when podcast_url is blank when trying to set
- [x] Wrap `controller.create` in a transaction; if feeder podcast create fails, rollback the distribution create

### DIFF
--- a/app/controllers/api/story_distributions_controller.rb
+++ b/app/controllers/api/story_distributions_controller.rb
@@ -13,6 +13,8 @@ class Api::StoryDistributionsController < Api::BaseController
 
   announce_actions resource: :story_resource
 
+  around_action :wrap_in_transaction, only: :create
+
   def story_resource
     resource.try(:story)
   end
--- a/config/initializers/faraday.rb
+++ b/config/initializers/faraday.rb
@@ -1,0 +1,2 @@
+require 'faraday'
+Faraday.default_adapter = :excon
--- a/test/models/distributions/podcast_distribution_test.rb
+++ b/test/models/distributions/podcast_distribution_test.rb
@@ -24,6 +24,16 @@ describe Distributions::PodcastDistribution do
   end
 
   it 'creates the podcast on feeder' do
+    distribution.stub(:get_account_token, 'token') do
+      distribution.url = nil
+      distribution.account.wont_be_nil
+      distribution.save!
+      distribution.distribute!
+      distribution.url.must_equal(podcast_url)
+    end
+  end
+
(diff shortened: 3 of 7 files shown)


kookster:
@gcampo88 how about taking a shot at reviewing this one?

kookster:
(ask me all the questions!)

kookster:
actually, I think I need to add this same logic to the episode level distro - hold off reviewing pls

gcampo88:
holding!

kookster:
ok - I think this is complete now @gcampo88

test/models/distributions/podcast_distribution_test.rb
```diff
@@ -7,19 +7,34 @@
...
+      distribution.account.wont_be_nil
+      distribution.save!
+      -> do
+        distribution.distribute!
+      end.must_raise(HyperResource::ServerError)
+    end
+  end
+
+  it 'throws error and rolls back if podcast on feeder fails' do
     distribution.stub(:get_account_token, 'token') do
```
gcampo88: seems like the name of this test & the one above should be swapped?

kookster: you're right! I changed the wrong one

config/initializers/faraday.rb
```diff
@@ -0,0 +1,2 @@
+require 'faraday'
+Faraday.default_adapter = :excon
```
cavis: What's this about?  Wasn't throwing errors the way your were expecting?

app/controllers/api/story_distributions_controller.rb
```diff
@@ -13,6 +13,8 @@ class Api::StoryDistributionsController < Api::BaseController
 
   announce_actions resource: :story_resource
 
+  around_action :wrap_in_transaction, only: :create
```
cavis: Fancy!

config/initializers/faraday.rb
kookster: excon is just better! no, really it makes it easier to debug consistently since then I can set ENV['EXCON_DEBUG'] = 'true' and both hyperresource and oauth2 gems will use excon and display the full logging of the requests and responses

app/controllers/api/story_distributions_controller.rb
kookster: (you know, back when I used to answer SO for fun.)

PR #211, opened 2016-12-20
Set the audio version template for a distribution

kookster:
- [x] Add in relationship between a distribution and an audio template
- [x] Show this link in representers for the distribution

This isn't validating audio by this, or specifying which audio to send to a distribution yet, need to handle those, esp. when we have multiple versions, but I believe that can wait for a later day.

### DIFF
--- a/app/models/distribution.rb
+++ b/app/models/distribution.rb
@@ -3,7 +3,10 @@ require 'hash_serializer'
 
 class Distribution < BaseModel
   belongs_to :distributable, polymorphic: true, touch: true
+  belongs_to :audio_version_template
+
   has_many :story_distributions
+
   serialize :properties, HashSerializer
 
   def distribute!
--- a/app/representers/api/distribution_representer.rb
+++ b/app/representers/api/distribution_representer.rb
@@ -17,4 +17,13 @@ class Api::DistributionRepresenter < Api::BaseRepresenter
       profile: model_uri(represented.owner)
     } if represented.id && represented.owner
   end
+
+  link rel: :audio_version_template, writeable: true do
+    {
+      href: api_audio_version_template_path(represented.audio_version_template)
+    } if represented.audio_version_template_id
+  end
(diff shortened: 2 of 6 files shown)


app/representers/api/distribution_representer.rb
```diff
@@ -17,4 +17,13 @@ def self_url(represented)
       profile: model_uri(represented.owner)
     } if represented.id && represented.owner
   end
+
+  link rel: :audio_version_template, writeable: true do
+    {
```
kookster: not doing it, sorry hound.

app/models/distribution.rb
```diff
@@ -3,7 +3,10 @@
 
 class Distribution < BaseModel
   belongs_to :distributable, polymorphic: true, touch: true
+  belongs_to :audio_version_template
```
cavis: I'm trying to think of a case where a distribution might have > 1 audio version .... but i can't.

cavis: Yeah, I think this is a pretty good assumption in our data universe.  Will also help to more explicitly define the purpose of an AudioVersion.

kookster: I may be wrong though, as on the b'cast side we do distribute via subauto multiple versions, like the single file version and a segmented version. So depending on how that gets implemented when we bring in b'cast, this could change. Currently that is modeled by having multiple templates associated to a single series via a series_id fk on the template table. I believe this could shift to associating the template to the form of distribution as well.

So that gets messy, it means a distro could have multiple templates. Likewise, I wonder if a template might make sense in multiple distros - that is fine with the current setup, but falls apart if we shift the fk to associate the template to the distro_id, at that point, I think we many need a join table :^/

kookster:
yeah, the required db change is in that prx.org PR https://github.com/PRX/prx.org/pull/287, so if and when we approve this, should run that first (esp. as this is an additive value).

PR #212, opened 2016-12-21
published dates in the future

kookster:
- [x] Allow published_at to be not null, but in the future, and not return the story as published until then
- [x] Update and retrieve published attribute respecting if the date is in the future

### DIFF
--- a/test/models/story_test.rb
+++ b/test/models/story_test.rb
@@ -173,11 +173,17 @@ describe Story do
     end
 
     it 'allows the published date to be set via boolean' do
-      story.published = false
-      story.published_at.must_be_nil
+      [false, 'f', 'false', '0', 0].each do |v|
+        story.published = v
+        story.published_at.must_be_nil
+        story.wont_be :published?
+      end
 
-      story.published = true
-      story.published_at.wont_be_nil
+      [true, 't', 'true', '1', 1].each do |v|
+        story.published = true
+        story.published_at.wont_be_nil
+        story.must_be :published?
+      end
     end
   end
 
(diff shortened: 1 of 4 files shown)


sandikbarr:
Should this be writeable in the representer? https://github.com/PRX/cms.prx.org/blob/master/app/representers/api/story_representer.rb#L11

test/models/story_test.rb
```diff
@@ -171,6 +171,14 @@
         story.unpublish!
       end.must_raise(RuntimeError)
     end
+
+    it 'allows the published date to be set via boolean' do
+      story.published = false
+      story.published.must_be_nil
```
gcampo88: did you mean for this (and ditto two lines down) to be `story.published_at`?

kookster: it may look that way, but I am overriding the `event_attribute` gem default functionality, and that means `published` and `published=` are affected: https://github.com/PRX/cms.prx.org/pull/212/files#diff-8a4ad5774f65a8a217b216fa2038a5bcR169

gcampo88: didn't you remove the `event_attribute :published_at` in this same PR? (I assumed this spec was to test the `published=` method, which takes care of setting `published_at` -- so wouldn't you want to check the `published at` attribute? Or am I confusing myself...)

kookster: yeah, I see what you mean, I'll make it better

test/models/story_test.rb
```diff
@@ -171,6 +171,20 @@
...
     end
+
+    it 'allows the published date to be set via boolean' do
+      [false, 'f', 'false', '0', 0].each do |v|
+        story.published = v
+        story.published_at.must_be_nil
+        story.wont_be :published?
+      end
+
+      [true, 't', 'true', '1', 1].each do |v|
```
kookster: fixed

PR #213, opened 2017-01-04
add in a release date used to set published_at on publish!

kookster:
Rather than directly updating the published date, that value is set when a story is published!

Usually, the published_at will be set to the current time when the story is published.

If a released_at value has been saved for the story, then when it is published!, the published_at is instead set to the published_at, be it past, present, or future.

Story is still only made published (set published_at to a value from nil) via the publish action on the story controller, and likewise unpublished (set published_at to nil) via a controller action.

### DIFF
--- a/app/models/story.rb
+++ b/app/models/story.rb
@@ -183,13 +183,12 @@ class Story < BaseModel
   end
 
   # raising exceptions here to prevent sending publish messages
-  #  when not actually a change to be published
-  def publish!(apublished_at = nil)
-    apublished_at = released_at || DateTime.now
+  # when not actually a change to be published
+  def publish!
     if !published_at.nil?
       raise "Story #{id} is already published."
     else
-      update_attributes!(published_at: apublished_at)
+      update_attributes!(published_at: (released_at || DateTime.now))
     end
   end
 
--- a/app/models/story.rb
+++ b/app/models/story.rb
@@ -51,6 +51,7 @@ class Story < BaseModel
   has_one :license, foreign_key: :piece_id
 
   before_validation :set_app_version, on: :create
+  before_validation :update_published_to_released
 
   # indicates piece is published with promos only - not full audio
   event_attribute :promos_only_at
@@ -64,7 +65,7 @@ class Story < BaseModel
   # indicates the piece is not publically available, only to the network
   event_attribute :network_only_at
 
-  scope :published, -> { where('`published_at` IS NOT NULL AND `published_at` <= now()') }
+  scope :published, -> { where('`published_at` <= now()') }
   scope :unpublished, -> { where('`published_at` IS NULL OR `published_at` > now()') }
   scope :unseries, -> { where('`series_id` IS NULL') }
   scope :v4, -> { where(app_version: PRX::APP_VERSION) }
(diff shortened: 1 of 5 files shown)


app/models/story.rb
```diff
@@ -183,7 +184,8 @@ def date_for_boolean(value)
 
   # raising exceptions here to prevent sending publish messages
   #  when not actually a change to be published
-  def publish!(apublished_at = DateTime.now)
+  def publish!(apublished_at = nil)
```
gcampo88: do we still need the parameter if it's being set in the next line no matter what's passed in? also, looks like it's never called with a param within CMS: 
screen shot 2017-01-05 at 11 12 08 am

app/models/story.rb
```diff
@@ -226,4 +228,14 @@ def set_app_version
     self.app_version = PRX::APP_VERSION
     self.deleted_at = DateTime.now
   end
+
+  # This is only if the story is published
+  # and the user wants to update the published_at value
+  # does not set the story published_at if it is not currently set to something
+  # will update the datetime when it is published_at
+  def update_published_to_released
+    if published_at && !released_at.nil? && released_at_changed?
```
gcampo88: ooh, where does `released_at_changed?` come from?

kookster: http://api.rubyonrails.org/classes/ActiveModel/Dirty.html

app/models/story.rb
kookster: yeah, good one - could get rid of this arg, especially now with `released_at` as the mechanism to set published_at

gcampo88: :+1:

app/models/story.rb
gcampo88: cool!

test/controllers/api/distributions_controller_test.rb
```diff
@@ -13,15 +13,20 @@
...
   describe '#create' do
+
+    before(:each) do
+      class << @controller; attr_accessor :prx_auth_token; end
+      @controller.prx_auth_token = token
+    end
+
     it 'can create a podcast distribution for a series' do
+      template_uri = "/api/v1/audio_version_templates/#{audio_version_template.id}"
+      podcast_uri = 'http://feeder.prx.org/api/v1/podcast/12345'
```
gcampo88: won't change anything, but do you want to use this `podcast_uri` variable on line 33 to avoid repeating the full uri there? (that line is untouched in this PR --         `res.update_attributes!(url: 'http://feeder.prx.org/api/v1/podcast/12345'`)

app/models/story.rb
```diff
@@ -64,7 +65,7 @@ def self.paranoia_scope
   # indicates the piece is not publically available, only to the network
   event_attribute :network_only_at
 
-  scope :published, -> { where('`published_at` IS NOT NULL AND `published_at` <= now()') }
+  scope :published, -> { where('`published_at` <= now()') }
```
cavis: 👍 ah, so `null < now()`

kookster: in mysql, if you are doing a comparison against a value, such as `<=`, then it will not match any records with a null value - not for `>` or `<`. So it is redundant in this case to exclude where it `is not null`, because by comparing against `now`, it already is.

app/models/story.rb
kookster: ok, latest commit should address

kookster:
@gcampo88 @cavis I think I have addressed the issue.

gcampo88:
👍

PR #214, opened 2017-01-05
Auto create story distros

kookster:
- [x] create story distributions when a story is created and is in a series that has distributions associated with it
- [x] send story attributes to feeder on episode create
- [x] include distributions in the story representer
- [x] bug fix: don't include stories in the series min representer - makes it huge

### DIFF
--- a/app/models/story.rb
+++ b/app/models/story.rb
@@ -47,13 +47,13 @@ class Story < BaseModel
   has_many :picks, foreign_key: 'playlistable_id'
   has_many :playlists, through: :picks
   has_many :purchases, foreign_key: :purchased_id
+  has_many :distributions, class_name: StoryDistribution, foreign_key: :piece_id
 
   has_one :promos, -> { where(promos: true) }, class_name: 'AudioVersion', foreign_key: :piece_id
   has_one :license, foreign_key: :piece_id
 
   before_validation :set_app_version, on: :create
   before_validation :update_published_to_released
-  after_commit :create_story_distributions, on: [:create]
 
   # indicates piece is published with promos only - not full audio
   event_attribute :promos_only_at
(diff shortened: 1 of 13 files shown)


app/representers/api/story_representer.rb
```diff
@@ -110,6 +110,17 @@ class Api::StoryRepresenter < Api::BaseRepresenter
     } if represented.id
   end
   embed :musical_works, paged: true, item_class: MusicalWork, zoom: false
+
+  link :distributions do
+    {
```
kookster: no, I like it how it is

test/models/story_test.rb
```diff
@@ -26,6 +26,29 @@
...
+    it 'creates story distributions' do
+      stub_request(:post, 'https://id.prx.org/token').
+        to_return(status: 200,
+                  body: '{"access_token":"abc123","token_type":"bearer"}',
+                  headers: { 'Content-Type' => 'application/json; charset=utf-8' })
+
+      stub_request(:get, 'https://feeder.prx.org/api/v1/podcasts/23').
+        to_return(status: 200, body: json_file(:podcast), headers: {})
+
+      stub_request(:post, 'https://feeder.prx.org/api/v1/podcasts/23/episodes').
```
cavis: Does webmock automatically assert that a stubbed request was actually made?

app/models/story.rb
```diff
@@ -212,6 +214,19 @@ def v4?
...
 
+  # for each distribution, create a story distribution
+  def create_story_distributions
+    return unless series
+    series.distributions.each do |distro|
+      begin
+        distro.create_story_distribution(self)
+      rescue StandardError => err
+        logger.error(err)
+        NewRelic::Agent.notice_error(err)
```
cavis: I like this.  Any case for also adding this to the series, when it's creating the feeder Podcast?

test/models/story_test.rb
kookster: you could say that - it throws an error so the test errors out if you don't stub the request.

app/models/story.rb
kookster: yes, we can expand using this - copying what you did in dovetail. I think if we want to do this more in the rails apps, we should probably abstract it, on the off chance that we quit using new relic, or so we can easily stub or change all error noticing

PR #217, opened 2017-01-06
Check if Series audio version template was deleted before linking to it for Story

gcampo88:
Handles what to do for child resources when Series is deleted, to nix 500 errors we're getting on Publish a) when you delete an Audio Version Template from a Series with existing Stories (solution: make the foreign_key of the audio version nil but keep the audio version), and b) when you delete a Series that has Distributions (solution: delete the Distributions too).

### DIFF
--- a/app/models/audio_version_template.rb
+++ b/app/models/audio_version_template.rb
@@ -5,7 +5,7 @@ class AudioVersionTemplate < BaseModel
 
   has_many :audio_file_templates, -> { order :position }, dependent: :destroy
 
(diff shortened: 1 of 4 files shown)


kookster:
So when deleting the template, it isn't nulling out the keys in the audio versions, so you end up with a template id in there for a deleted template?

I think that ActiveRecord should be able to null those out - curious as to if this case is preventable.

gcampo88:
good call @kookster ! Back to you

PR #224, opened 2017-01-12
Fix API responses

kookster:
- [x] On updates (`PUT`), return the updated resource in the body, as with `POST` responses
- [x] Fix `DELETE` responses to correctly return an error when one is thrown, e.g. on 404
- [x] Fix authorization error response to match `hal_api-rails` errors, with `:status`, `:message`, and `:backtrace` attributes

### DIFF
--- a/app/controllers/api/api_responder.rb
+++ b/app/controllers/api/api_responder.rb
@@ -1,4 +1,4 @@
-require "roar/rails/responder"
+require 'roar/rails/responder'
 
 class Api::ApiResponder < Roar::Rails::Responder
   def api_behavior
--- a/app/controllers/api/api_responder.rb
+++ b/app/controllers/api/api_responder.rb
@@ -1,0 +1,17 @@
+require "roar/rails/responder"
+
+class Api::ApiResponder < Roar::Rails::Responder
+  def api_behavior
+    raise MissingRenderer.new(format) unless has_renderer?
+
+    if post?
+      display(resource, status: :created)
+    elsif put?
+      display(resource, status: :ok)
+    elsif delete?
+      display(resource, status: :no_content)
+    else
+      super
+    end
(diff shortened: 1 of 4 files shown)


kookster:
@sandikbarr @cavis pls review, this is the one that returns a resource on `PUT` and so has implications for `publish`

app/controllers/api/api_responder.rb
```diff
@@ -0,0 +1,17 @@
...
+class Api::ApiResponder < Roar::Rails::Responder
+  def api_behavior
+    raise MissingRenderer.new(format) unless has_renderer?
+
+    if post?
+      display(resource, status: :created)
+    elsif put?
+      display(resource, status: :ok)
+    elsif delete?
+      display(resource, status: :no_content)
```
sandikbarr: Does this mean the resource is returned for DELETE too but with a 204?

kookster: good question.

no, when you tell it to do a 204 (`status: :no_content`), it is smart enough in that display method to know there will be no body.

However, by calling display, it is then smart enough to actually return with errors that get thrown in the action processing. The default before this was to call a `head :no_content`, and that didn't.

app/controllers/api/api_responder.rb
```diff
@@ -0,0 +1,17 @@
+require 'roar/rails/responder'
+
+class Api::ApiResponder < Roar::Rails::Responder
+  def api_behavior
+    raise MissingRenderer.new(format) unless has_renderer?
+
+    if post?
+      display(resource, status: :created)
+    elsif put?
+      display(resource, status: :ok)
```
cavis: Worth noting that getting a 200 for a PUT is going to break some parts of Publish, which explicitly expects the 204.

Issue #231, opened 2017-01-18
Deleting a Series should cascade down to its children


gcampo88:
Otherwise, orphaned stories will yield: {"status":500,"message":"No route matches {:action=>\"show\", :api_version=>\"v1\", :controller=>\"api/series\", :id=>nil} missing required keys: [:id]"}.

gcampo88:
handled in #232.

PR #232, opened 2017-01-18
Nullify fk to series in child stories when parent series is deleted

gcampo88:
#231


kookster:
reopening - we'll do a nullify instead of destroy

PR #239, opened 2017-01-30
Podcast import

kookster:
Okay, we'll be serving podcast imports 2 ways:
- [x] `lib/podcast_importer.rb` can import using only API calls, can be run from anywhere
- [x] add an authorized podcast_import endpoint, and model to track the import, which can be integrated into publish

Requires PRX/prx.org#291


lib/itunes_category_validator.rb
```diff
@@ -0,0 +1,60 @@
...
+  def self.category?(cat)
+    CATEGORIES.keys.include?(cat)
+  end
+
+  def category?(cat)
+    self.class.category?(cat)
+  end
+
+  def self.subcategory?(subcat, category = nil)
+    cat ||= category
```
gcampo88: why overwrite `category`?

app/models/podcast_import.rb
```diff
@@ -0,0 +1,267 @@
...
+    if !feed.itunes_image.blank?
+      series.images.create!(
+        upload: feed.itunes_image,
+        purpose: Image::PROFILE
+      )
+    end
+
+    if feed.image && feed.image.url
+      series.images.create!(
+        upload: feed.image.url,
```
gcampo88: Perhaps-uninformed question: how come here the upload points to the url, and for itunes image it points to the itunes image as a whole?

app/models/story_distributions/episode_distribution.rb
```diff
@@ -10,13 +10,26 @@ def distribute!
...
     podcast = distribution.get_podcast
-    episode = podcast.episodes.post(episode_attributes)
+    episode = podcast.episodes.post(episode_attributes.merge(create_attributes))
     episode_url = URI.join(feeder_root, episode.links['self'].href).to_s
     raise 'Failed to get episode url on create' if episode_url.blank?
     update_attributes!(url: episode_url) if episode_url
+    episode
+  end
+
+  def get_episode
```
gcampo88: are you using this anywhere? Or just for future convenience?

app/representers/api/auth/account_representer.rb
```diff
@@ -15,6 +15,19 @@ class Api::Auth::AccountRepresenter < Api::AccountRepresenter
...
+    {
+      href: "#{api_authorization_account_podcast_imports_path(represented)}#{index_url_params}",
+      templated: true,
+      count: represented.podcast_imports.count
+    }
+  end
+  embed :podcast_imports,
+        paged: true,
+        item_class: PodcastImport,
+        item_decorator: Api::Auth::StoryMinRepresenter,
```
gcampo88: not Api::Auth::PodcastImportRepresenter ?

lib/feeder_importer.rb
```diff
@@ -0,0 +1,72 @@
...
+      create_episodes(response['_embedded']['prx:items'])
+    end
+  end
+
+  def create_stories(episodes)
+    episodes.each { |e| create_story(e) }
+  end
+
+  def create_story(episode)
+    # create the story
```
gcampo88: placeholder?

lib/podcast_importer.rb
```diff
@@ -0,0 +1,267 @@
+# import script for a podcast live in feeder & dovetail
+require 'prx_access'
+require 'addressable/uri'
+require 'feedjira'
+require 'itunes_category_validator'
+
+class PodcastImporter
```
gcampo88: So much similarity between this & PodcastImport -- also, both this and FeederImporter say `# import script for a podcast live in feeder & dovetail` at top. Wondering when each different one will be used?

test/lib/feeder_importer_test.rb
```diff
@@ -0,0 +1,68 @@
...
+      with(headers: { 'Authorization' => 'Bearer thisisnotatoken' }).
+      to_return(status: 200, body: json_file('distributions_empty'), headers: {})
+
+    stub_request(:post, 'http://cms.prx.org/api/v1/series/12345/distributions').
+      with(body: '{"kind":"podcast","url":"http://feeder.prx.tech/api/v1/podcasts/23"}',
+           headers: { 'Authorization' => 'Bearer thisisnotatoken' }).
+      to_return(status: 200, body: json_file('podcast_distribution'), headers: {})
+  end
+
+  # it 'importer has account and feed' do
```
gcampo88: all commented out?

app/controllers/api/auth/podcast_imports_controller.rb
```diff
@@ -0,0 +1,20 @@
...
+  include ApiAuthenticated
+
+  api_versions :v1
+
+  filter_resources_by :account_id, :series_id
+
+  filter_params :url
+
+  def after_create_resource(res)
+    res.try(:import)
```
cavis: So the API create will result in a synchronous import running?  How fast can this run?  (I'm all for sync operations, by the way!)

Also, I'm not sure what happens when an error bubbles up to this after_create hook.  Do you still get a 201?

app/models/podcast_import.rb
```diff
@@ -0,0 +1,267 @@
...
+    create_attributes[:keywords] = (entry[:itunes_keywords] || '').split(',').map(&:strip)
+    create_attributes[:position] = entry[:itunes_order]
+    create_attributes[:url] = episode_url(entry)
+
+    distro.add_episode_to_feeder(create_attributes)
+  end
+
+  def episode_url(entry)
+    url = entry[:feedburner_orig_link] || entry[:url] || entry[:link]
+    if url =~ /libsyn\.com/
```
cavis: Not a fan of libsyn, huh?

app/representers/api/auth/podcast_import_representer.rb
```diff
@@ -0,0 +1,28 @@
...
+  link rel: :account, writeable: true do
+    {
+      href: api_account_path(represented.account),
+      title: represented.account.name,
+      profile: model_uri(represented.account)
+    } if represented.id && represented.account_id
+  end
+  embed :account, class: Account, decorator: Api::Min::AccountRepresenter
+
+  link rel: :series, writeable: true do
```
cavis: Not sure why this is writeable - i thought the series was created for you.  But I'll keep reading!

lib/feeder_importer.rb
```diff
@@ -0,0 +1,72 @@
+# import script for a podcast live in feeder & dovetail
+require 'prx_access'
+
+class FeederImporter
```
cavis: So this is pretty much a standalone script at the moment, right?  (Not accessible via the api)

lib/feedjira/parser/podcast.rb
```diff
@@ -0,0 +1,82 @@
+require 'feedjira'
+require 'feedjira/parser/media_content'
+require 'feedjira/parser/media_group'
+require 'feedjira/parser/podcast_image'
+require 'feedjira/parser/podcast_item_enclosure'
+require 'feedjira/parser/podcast_item'
+
+module Feedjira
+  module Parser
+    class Podcast
```
cavis: This is all in feeder as well, correct?  Should probably make this a shared module at some point.

lib/feeder_importer.rb
cavis: Yeah, I'm a bit confused why when one would use these standalone libs to run an import against an existing API ... vs just using the CMS API (which can handle this on its own via the PodcastImport model).

lib/podcast_importer.rb
cavis: That, and (per my other comment) when would these lib importers be used vs the model importer?

test/lib/feeder_importer_test.rb
kookster: yeah, started on this, but not going to get it done in this PR

lib/itunes_category_validator.rb
kookster: In my defense - I didn't write this, brought it over from feeder which has the same logic.
perhaps someone was concerned about overwriting an arg?
in any case, I'll fix this, and the `each` below it that should be a `detect`...

app/models/podcast_import.rb
kookster: yes, what @cavis said - they are structured differently in the xml, and so have to be accessed differently from the feedjira parse of the feed.

app/models/story_distributions/episode_distribution.rb
kookster: I was using this until I refactored to do a single create rather than a create then update for the feeder episode.

I could delete it, but it is convenient, and mirrors `get_podcast` from the `PodcastDistribution`

app/representers/api/auth/account_representer.rb
kookster: oops. fixing.

lib/feeder_importer.rb
kookster: you can ignore this, not finished - I started working on it, but then changed to the podcast importer b/c we need it first.

lib/podcast_importer.rb
kookster: I wrote the `lib/podcast_importer.rb` first, thinking I could do everything with API calls so that the importer could be run anywhere - e.g. you can run it on your local machine agains the prod cms and feeder APIs to import a podcast.

...but it was problematic.
First off - slow to hit the API for everything.
Lots of creates then having to do updates and publish requests after.
Couldn't directly set the published_at date to match the feed.

And basically not useful to integrate with publish.
If we want this to work with publish, needed an API endpoint, and to do the processing async.
But if it is async, you need a model to keep track of how the import is going.

So, I created a podcast import model to both keep track of the progress of the import, and to be accessible via the API to track that progress.

In short, the lib importer is good for importing test data, or other dev tasks, and honestly I spent a long time on it, and I wasn't going to delete it.

The model importer is meant to be integrated into publish so a user could start an import prpcess of their existing podcast.

app/models/podcast_import.rb
kookster: libsyn sets the `link` to the same value as the mp3 enclosure on libsyn.
We don't want to keep linking to libsyn after importing the feed, so I update the url in that special case.

app/controllers/api/auth/podcast_imports_controller.rb
kookster: yeah, I changed method names, that should be `import_later`, fixing.

app/representers/api/auth/podcast_import_representer.rb
kookster: I was envisioning we might want someone to be able to import into an existing series, but yeah, could be readonly.

lib/feeder_importer.rb
kookster: see long note above. tl;dr made the standalone first, then optimized it to create the api importer.

I can imagine cases where I might want to be able to import from my local to some test/prod system in a synch way.

lib/feedjira/parser/podcast.rb
kookster: yes, several libs should be, like prx_access and this, perhaps the itunes categories stuff too

cavis:
👍  all good for me here.

gcampo88:
LGTM!

Issue #216, opened 2017-01-06
Use the list of accounts on the token for authorization, not all those associated with the user


kookster:
- [x] Get the list of accounts the token is authorized for from the `aur` hash/list in the token
- [x] Handle the different cases for how `aur` can express lists of accounts, I think the rack gem handles most of this: https://github.com/PRX/rack-prx_auth/blob/master/lib/rack/prx_auth/token_data.rb#L32
- [x] Use this list of account ids (any role) instead of querying memberships to determine accounts to use in the authorization controller

kookster:
I am behind on my other tix, so reassigning to @gcampo88

gcampo88:
handled in #240

PR #240, opened 2017-01-31
Use token list of accounts for user authorization

gcampo88:
#216
- extracts account authorization information from the auth token in the ApiAuthenticated concern
- uses that list of account IDs instead of querying the memberships table to determine which accounts to use 
- updates StubToken class in `test_helper.rb` to more closely  mimic auth token behavior in terms of authorized_resources list

### DIFF
--- a/app/controllers/api/auth/stories_controller.rb
+++ b/app/controllers/api/auth/stories_controller.rb
@@ -38,7 +38,7 @@ class Api::Auth::StoriesController < Api::StoriesController
     @stories ||= if params[:network_id]
       super.published
     else
-      Authorization.new(prx_auth_token).token_auth_stories
+      authorization.token_auth_stories
     end
   end
 
@@ -46,8 +46,8 @@ class Api::Auth::StoriesController < Api::StoriesController
     super.tap do |story|
       story.creator_id = current_user.id
       story.account_id ||= story.series.try(:account_id)
-      story.account_id ||= current_user.account_id
-      story.account_id ||= current_user.approved_accounts.first.try(:id) # not sure if I should change this to look at token
+      story.account_id ||= current_user.account_id if authorization.authorized?(current_user.default_account)
+      story.account_id ||= authorization.token_auth_accounts.first.try(:id)
     end
   end
 end
--- a/app/controllers/api/auth/stories_controller.rb
+++ b/app/controllers/api/auth/stories_controller.rb
@@ -38,7 +38,7 @@ class Api::Auth::StoriesController < Api::StoriesController
     @stories ||= if params[:network_id]
       super.published
     else
-      current_user.approved_account_stories
+      Authorization.new(prx_auth_token).token_auth_stories
     end
   end
 
@@ -47,7 +47,7 @@ class Api::Auth::StoriesController < Api::StoriesController
       story.creator_id = current_user.id
       story.account_id ||= story.series.try(:account_id)
       story.account_id ||= current_user.account_id
-      story.account_id ||= current_user.approved_accounts.first.try(:id)
+      story.account_id ||= current_user.approved_accounts.first.try(:id) # not sure if I should change this to look at token
     end
   end
 end
--- a/app/controllers/concerns/api_authenticated.rb
+++ b/app/controllers/concerns/api_authenticated.rb
@@ -7,18 +7,11 @@ module ApiAuthenticated
   extend ActiveSupport::Concern
 
   included do
-    before_filter :authenticate_user!, :get_authorized_resources
+    before_filter :authenticate_user!
   end
 
   def authenticate_user!
     user_not_authorized unless current_user
-  end
-
-  def get_authorized_resources
-    token_accounts = prx_auth_token.authorized_resources.try(:keys)
-    if token_accounts
-      current_user.approved_accounts ||= Account.where(id: token_accounts)
-    end
   end
 
   def cache_show?
--- a/app/controllers/concerns/api_authenticated.rb
+++ b/app/controllers/concerns/api_authenticated.rb
@@ -14,6 +14,10 @@ module ApiAuthenticated
     user_not_authorized unless current_user
   end
 
+  def authorization
+    Authorization.new(prx_auth_token)
+  end
+
   def cache_show?
     false
   end
--- a/app/controllers/concerns/api_authenticated.rb
+++ b/app/controllers/concerns/api_authenticated.rb
@@ -7,11 +7,17 @@ module ApiAuthenticated
   extend ActiveSupport::Concern
 
   included do
-    before_filter :authenticate_user!
+    before_filter :authenticate_user!, :get_authorized_resources
   end
 
   def authenticate_user!
     user_not_authorized unless current_user
+  end
+
+  def get_authorized_resources
+    if prx_auth_token.authorized_resources
+      current_user.approved_accounts ||= Account.where({ id: prx_auth_token.authorized_resources.keys })
+    end
   end
 
   def cache_show?
--- a/app/controllers/concerns/api_authenticated.rb
+++ b/app/controllers/concerns/api_authenticated.rb
@@ -15,8 +15,9 @@ module ApiAuthenticated
   end
 
   def get_authorized_resources
-    if prx_auth_token.authorized_resources
-      current_user.approved_accounts ||= Account.where({ id: prx_auth_token.authorized_resources.keys })
+    token_accounts = prx_auth_token.authorized_resources.try(:keys)
+    if token_accounts
+      current_user.approved_accounts ||= Account.where(id: token_accounts)
     end
   end
 
--- a/app/models/authorization.rb
+++ b/app/models/authorization.rb
@@ -6,10 +6,6 @@ class Authorization
 
   def initialize(token)
     @token = token
-    get_authorized_resources
-  end
-
-  def get_authorized_resources
   end
 
   def token_auth_accounts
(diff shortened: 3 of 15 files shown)


app/controllers/concerns/api_authenticated.rb
```diff
@@ -7,13 +7,20 @@ module ApiAuthenticated
...
   end
 
   def authenticate_user!
     user_not_authorized unless current_user
   end
 
+  def get_authorized_resources
+    token_accounts = prx_auth_token.authorized_resources.try(:keys)
+    if token_accounts
+      current_user.approved_accounts ||= Account.where(id: token_accounts)
```
kookster: I'm not sure we want the user model, which is retrieved from the DB and is in the model layer, also aware of the token and its contents which come from the controller layer. Seems like we are having controller / request logic and information leak into the model which should not be aware of or rely on that layer.

The other part of this is that it is could be misleading, because the user may have a different set of accounts they can access than is allowed by and listed in the token.

For example, you could generate a token with only a subset of the actual accounts the user has access to, or using client credentials, get a different account entirely in `aur` than is reflected in the user's account membership.

So the more I think about it, it seems like with the move to get authorized accounts from the token, not the user, they become 2 different things - there is the user that logged in, and the list of accounts authorized in the token, and they are not necessarily related except for having both been identified in the token.

app/models/user.rb
```diff
@@ -45,25 +47,15 @@ def name
...
   def approved_active_accounts
-    approved_accounts.active
+    approved_accounts.try(:active)
   end
 
   def approved_account_stories
-    Story.
-      joins('LEFT OUTER JOIN `memberships` ON `memberships`.`account_id` = `pieces`.`account_id`').
-      where(['memberships.user_id = ? and memberships.approved is true', id])
+    !approved_accounts.nil? ? Story.where(account_id: approved_accounts.try(:ids)) : []
```
kookster: could `Story.where(account_id: approved_accounts.try(:ids))`  be in a controller method instead of in the user model? doesn't seem to rely be using anything in the user model, except the convenience that it is the `resource` loaded by authorized controllers.

app/controllers/concerns/api_authenticated.rb
kookster: Curious what @cavis might say about this as well - with the separation of accounts and users, where the token provides the list of accounts not the user, I think it makes sense to move these methods using the list of accounts out of the user model, perhaps into methods on the `ApiAuthenticated` concern, or just in the relevant controller?

app/models/user.rb
gcampo88: Could certainly be in there! I'm not quite sure where to place the method definitions for `approved_account_series` and `approved_account_stories` because they are called in the representers, which would lead me to want to put them in the model layer, but also are called in the controller...

kookster: yeah, that is annoying - looks like they do get used for counts in the authorization_representer, but only that one representer at least.

kookster: We can talk about it when I get in - one way to handle this may be to create an `authorization` resource based on the token, which could encompass both access to the user, and the list of accounts. That's sorta what feeder does, if you have a look at that, where the auth controller has this for the resource:
```
  def resource
    Authorization.new(prx_auth_token)
  end
```
And then the authorization model is basically a wrapper around the token, accessing the related user(well, user_id, feeder would have to make a remote call to getthe user), and could also return the approved_accounts and such:

```
# encoding: utf-8

class Authorization
  include HalApi::RepresentedModel

  attr_accessor :token

  def initialize(token)
    @token = token
  end

  def user_id
    token.user_id
  end
...
end
```

kookster: That would separate the idea of what info comes from the token and it's list of accounts vs. the user and the list of all member accounts, might be better?

app/controllers/api/auth/stories_controller.rb
```diff
@@ -47,7 +47,7 @@ def create_resource
       story.creator_id = current_user.id
       story.account_id ||= story.series.try(:account_id)
       story.account_id ||= current_user.account_id
-      story.account_id ||= current_user.approved_accounts.first.try(:id)
+      story.account_id ||= current_user.approved_accounts.first.try(:id) # not sure if I should change this to look at token
```
gcampo88: This is in the stories controller -- not sure whether to change this to just use the first account ID listed on the auth token, or perhaps the first account ID with role "admin", or else to leave it as is -- calling the User model, which joins against memberships. (Or should this simply be the user's `individual_account` ? )

gcampo88:
@kookster this more along the lines of what you're thinking?

app/controllers/api/auth/stories_controller.rb
kookster: So this is the fallback for the fallback (when there is neither a series.account nor a default user account), so I think it's fine to fallback on the first account, creating a story doesn't require :admin permissions.

There is another issue here though which is that the default account or series account may not be in the token's list of accounts, so we should check for that too - we shouldn't set the account to a value that the token does not include, does that make sense?

app/models/authorization.rb
```diff
@@ -0,0 +1,23 @@
+# encoding: utf-8
+
+class Authorization
```
kookster: I like this whole class.

test/models/user_test.rb
```diff
@@ -39,9 +41,8 @@
     user.networks.must_include network
   end
 
-  it 'has a list of stories for approved accounts' do
-    start_count = user.individual_account.stories.count
-    create(:account, user: user, stories_count: 2)
-    user.approved_account_stories.count.must_equal start_count + 2
+  it 'has a list of stories and series' do
+    user.approved_account_stories.must_include story
```
kookster: Are you leaving the User methods like `approved_account_stories`, even though they aren't being used anymore?

gcampo88: `approved_account_stories` and `approved_account_series` are both used in the Authorization Representer. Do you think that representer should be looking at the stories + series associated with the accounts on the token instead?

app/models/authorization.rb
```diff
@@ -0,0 +1,35 @@
...
+  attr_accessor :token
+  attr_reader :token_auth_accounts
+
+  def initialize(token)
+    @token = token
+  end
+
+  def id
```
gcampo88: I added this because the AuthorizationRepresenter wants an id to build links ... misguided?

app/models/authorization.rb
```diff
@@ -0,0 +1,35 @@
...
+    default_account.id
+  end
+
+  def default_account
```
gcampo88: similarly, added this because AuthorizationRepresenter wants to return a default_account

gcampo88:
ok, back to you @kookster !

app/controllers/api/auth/stories_controller.rb
```diff
@@ -38,16 +38,16 @@ def resources_base
...
     end
   end
 
   def create_resource
     super.tap do |story|
       story.creator_id = current_user.id
       story.account_id ||= story.series.try(:account_id)
-      story.account_id ||= current_user.account_id
-      story.account_id ||= current_user.approved_accounts.first.try(:id)
+      story.account_id ||= current_user.account_id if authorization.authorized?(current_user.default_account)
```
kookster: thanks for adding that check - clever to use the `authorized?` method, I was thinking of doing a detect on the account list - this is much better

app/models/authorization.rb
kookster: what does it want the id for? the `self` link? I don't think you should need to provide an id for the representer, can we track this down?

kookster: if it is the `self` link, which is all I can think of, we can override the `self_link` method on the representer to not need this.

app/models/authorization.rb
kookster: yeah, makes sense to me

app/models/authorization.rb
```diff
@@ -0,0 +1,35 @@
...
+    User.find(token.user_id).default_account
+  end
+
+  def token_auth_accounts
+    token_ids = token.authorized_resources.try(:keys)
+    @token_auth_accounts = Account.where(id: token_ids) if token_ids
+  end
+
+  def token_auth_stories
+    Story.where(account_id: token_auth_accounts.try(:ids)) unless token_auth_accounts.nil?
```
kookster: do you need the `try` when you have the `unless nil?` already? Fine to leave it this way, I am a belt and suspenders kind of person myself.

test/models/authorization_test.rb
```diff
@@ -0,0 +1,26 @@
...
+    authorization.token_auth_accounts.must_include account
+  end
+
+  it 'has lists of stories and series belonging to accounts authorized on token' do
+    authorization.token_auth_stories.must_equal account.stories
+    authorization.token_auth_series.must_equal account.series
+  end
+
+  it 'checks against token to see if accounts are authorized' do
+    authorization.authorized?(account).must_equal true
```
kookster: I believe you could also write this as:
  `authorization.must_be :authorized?, account`

just a point of information - this test is great

app/models/authorization.rb
gcampo88: just doing belt + suspenders!

app/models/authorization.rb
gcampo88: strangely, I think it's the `to_json` method. Looks like ActiveRecord has its own `to_json`, and I think since Authorizations are not db objects and therefore have no IDs, it's erroring out on `let(:json) { JSON.parse(representer.to_json) }`

kookster: bug in hal gem - https://github.com/PRX/hal_api-rails/issues/8

Issue #243, opened 2017-02-01
Announce series updates when images are updated


kookster:

gcampo88:
Did a little exploring on this.. is this the kind of thing we're thinking here?:
```Ruby
class Api::SeriesImagesController < Api::BaseController
  announce_actions :create, :destroy, decorator: Api::Msg::ImageRepresenter, subject: :image
  announce_actions :update, resource: :series_resource
end
```

gcampo88:
needs #258

PR #250, opened 2017-02-08
WIP announce things

gcampo88:
Is this the right way to go for #243?

### DIFF
--- a/app/controllers/api/series_images_controller.rb
+++ b/app/controllers/api/series_images_controller.rb
@@ -5,7 +5,12 @@ class Api::SeriesImagesController < Api::BaseController
 
   filter_resources_by :series_id
 
-  announce_actions decorator: Api::Msg::ImageRepresenter, subject: :image
+  announce_actions :create, :destroy, decorator: Api::Msg::ImageRepresenter, subject: :image
+  announce_actions :update, resource: :series_resource
 
   represent_with Api::ImageRepresenter
+
+  def series_resource
+    resource.try(:series)
+  end
 end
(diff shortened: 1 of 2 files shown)


cavis:
@gcampo88 yeah, i think that's generally the way we announce on the parent resource.  Do want to make sure the image is still announcing itself for create/update/delete, as the image-lambda relies on those.  (Looks like you may have removed "update").

BUT... I do think we need to have a larger conversation on how worker-callback announcements will work (per #255).  Because what will happen is:

1. Image is POSTed - series announces itself, but since the image is unprocessed, feeder doesn't accept it yet.
2. Lambda gets notified by the image announcement - process image
3. CMS worker gets callback from lambda, changes the image status to COMPLETE
4. Nothing happens - feeder doesn't get notified of this change, since currently the work has no means to trigger notifications.

gcampo88:
@cavis yup, agreed. So for now, all announcing happens through controllers, right? Do we want to do something like `Announce.publish(:audio, :publish, id: audio.id)` at the end of the worker? Or are you thinking we should make Announce figure in to the models?

gcampo88:
also @cavis, re: "Do want to make sure the image is still announcing itself for create/update/delete, as the image-lambda relies on those. (Looks like you may have removed "update")."
So we can announce the same action on multiple resources? I thought it must be one only. Could we do this?:
```
announce_actions :create, :update, :destroy, decorator: Api::Msg::ImageRepresenter, subject: :image
announce_actions :update, resource: :series_resource
```

cavis:
I think you can get this PR done, but it may not actually fix the issue.  (Since it'll announce the series to feeder, but before the image is ready).  Per #255 - need to have that conversation on how to announce outside of the controllers.

> So we can announce the same action on multiple resources?

I guess I don't know.  If we can't, that's a problem.  Because we definitely both here ... any image create/update/delete should announce both the image (for lambda) and the series (for feeder).

PR #251, opened 2017-02-09
Handle changing series accounts

### LINKED ISSUE #247: Update episodes of a series when series account changes
If a series is updated, and moved to a different account, also update all the episodes within that series to be a member of the same account.

### LINKED ISSUE #252: Allow the account for a series to be changed
Change the account link in the series representer to be `writeable`.
  https://github.com/PRX/cms.prx.org/blob/master/app/representers/api/series_representer.rb#L44

We have this sneaky trick to designate links in the representer as `writeable`:
  https://github.com/PRX/hal_api-rails/blob/master/lib/hal_api/representer/link_serialize.rb#L9-L25

This adds a not-so-obvious `set_[name]_uri` property that will update the relation.

Here's an example link where this is used to make the account writeable on a story:
  https://github.com/PRX/cms.prx.org/blob/master/app/representers/api/story_representer.rb#L44

gcampo88:
Fixes #252 (allowing series account to be changed) and fixes #247 (updating episodes in a series when series account changes)

### DIFF
--- a/app/models/series.rb
+++ b/app/models/series.rb
@@ -153,6 +153,6 @@ class Series < BaseModel
   end
 
   def update_account_for_stories
-    stories.each {|s| s.update_attributes!(account_id: account_id)} if account_id_changed?
+    stories.each { |s| s.update_attributes!(account_id: account_id) } if account_id_changed?
   end
 end
--- a/app/models/series.rb
+++ b/app/models/series.rb
@@ -37,6 +37,7 @@ class Series < BaseModel
   has_many :images, -> { where(parent_id: nil) }, class_name: 'SeriesImage', dependent: :destroy
 
   before_validation :set_app_version, on: :create
+  before_validation :update_account_for_stories, on: :update
 
   event_attribute :subscriber_only_at
 
@@ -149,5 +150,9 @@ class Series < BaseModel
   def set_app_version
     return unless new_record?
     self.app_version = PRX::APP_VERSION
+  end
+
+  def update_account_for_stories
+    stories.each {|s| s.update_attributes!(account_id: account_id)} if account_id_changed?
   end
 end
--- a/app/models/series.rb
+++ b/app/models/series.rb
@@ -37,7 +37,7 @@ class Series < BaseModel
   has_many :images, -> { where(parent_id: nil) }, class_name: 'SeriesImage', dependent: :destroy
 
   before_validation :set_app_version, on: :create
-  before_validation :update_account_for_stories, on: :update
+  after_save :update_account_for_stories, on: :update
 
   event_attribute :subscriber_only_at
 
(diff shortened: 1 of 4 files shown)


app/models/series.rb
```diff
@@ -37,6 +37,7 @@ def description_md
   has_many :images, -> { where(parent_id: nil) }, class_name: 'SeriesImage', dependent: :destroy
 
   before_validation :set_app_version, on: :create
+  before_validation :update_account_for_stories, on: :update
```
kookster: I tend not to make changes to child models in `before` operations. `before_validation` will run even if the change to the series doesn't save, or if `valid?` is called on the model and no save will occur. I think I would do this either in `after_save`, or `after_commit`

app/models/series.rb
```diff
@@ -150,4 +151,8 @@ def set_app_version
     return unless new_record?
     self.app_version = PRX::APP_VERSION
   end
+
+  def update_account_for_stories
+    stories.each { |s| s.update_attributes!(account_id: account_id) } if account_id_changed?
```
kookster: I am a fan of using the bang operation for update_attributes, especially in this kind of case where it will kick out of the each loop, and rollback the save 👍

app/models/series.rb
kookster: On a related note, if we use `after_commit`, we should make sure that callback will raise an exception: http://edgeguides.rubyonrails.org/upgrading_ruby_on_rails.html#error-handling-in-transaction-callbacks

kookster: yup, we have `raise_in_transactional_callbacks` set to true 👍

app/representers/api/series_representer.rb
```diff
@@ -41,7 +41,7 @@ class Api::SeriesRepresenter < Api::BaseRepresenter
   end
   embed :images, paged: true, item_class: SeriesImage, item_decorator: Api::ImageRepresenter
 
-  link :account do
+  link rel: :account, writeable: true do
```
kookster: exactly.

test/representers/api/series_representer_test.rb
```diff
@@ -21,4 +21,13 @@
...
     json['_links']['prx:audio-version-templates'].wont_be_nil
   end
+
+  it 'can set the account' do
+    series.account_id.wont_be_nil
+    series_hash = { title: 'Title', set_account_uri: 'api/v1/accounts/123' }
+    s_representer = Api::SeriesRepresenter.new(Series.new)
+    s_series = s_representer.from_json(series_hash.to_json)
+    s_series.title.must_equal 'Title'
+    s_series.account_id.must_equal 123
```
kookster: good test :dancing_men:

app/models/series.rb
gcampo88: :+1:

app/models/series.rb
gcampo88: 👍

gcampo88:
good call @kookster. back to you!

PR #254, opened 2017-02-14
Set story or series account on create

kookster:
We can call create and other authorized operations on the standard api controllers, not only the ones  under `\auth`.

This PR moves all the logic for setting accounts for stories and series on create to the standard api controllers, which also work for the child `auth` controllers.

It also moves the `authorization` method to the base controller, since that can be used on any api controller, not just under auth.

### DIFF
--- a/test/controllers/api/stories_controller_test.rb
+++ b/test/controllers/api/stories_controller_test.rb
@@ -17,6 +17,9 @@ describe Api::StoriesController do
       class << @controller; attr_accessor :prx_auth_token; end
       @controller.prx_auth_token = token
       @request.env['CONTENT_TYPE'] = 'application/json'
+
+      # for `create`, feeder calls don't need to work, but webmock needs a response defined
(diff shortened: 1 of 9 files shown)


test/controllers/api/stories_controller_test.rb
```diff
@@ -35,8 +35,30 @@ class << @controller; attr_accessor :prx_auth_token; end
       Story.find(id).account_id.must_equal account.id
     end
 
+    it 'can create a new story for a series' do
+      stub_request(:post, 'https://id.prx.org/token').to_return(status: 500)
```
gcampo88: why 500?

gcampo88: oh never mind, just saw the comment from earlier (might be worth reinstating?) 
`# feeder calls do't need to work, but webmock needs a response defined`

gcampo88: also could be worth putting this in the before(:each) ?

kookster:
good feedback @gcampo88 - done.

PR #255, opened 2017-02-15
Mark story as valid/invalid based on versions & files complying with templates

### LINKED ISSUE #148: audio file/version validations
- [x] Validate stories based on their series audio file/version templates
  - And for any story, lack of audio == invalid
- [x] Validate audio segments/durations based on templates
  - Validation won't actually happen until the cms-audio-lambda callback
  - But assume invalid until the callback is received (story will be "invalid" until all callbacks finish)
- [x] Add a top-level valid/invalid flag and reason on the story representer
- [x] Confirm that all story changes (include audio file/version changes) are being announced
  - So that feeder can pick up the changes, and publish a new podcast version whenever the valid flag flips
  - Feeder is the gatekeeper for valid/invalid stories - always keep the last valid state

gcampo88:
Fixes #148:

- Validates audio files' compliance with audio file template (if exists)
- validates audio versions by their audio files' validity + their own compliance with audio version template
- validates stories by their audio versions' validity + presence of audio. 
  - Does this all in `before_save` hooks, so will all be triggered by the `audio.save!` line in the audio callback worker.
  - After that `save!`, `announce`s the story update from the audio callback worker after audio file change. 
- Adds status & status_msg to representers for stories, audio versions, and audio files

Misc:
- adds a few concerns for DRYness and convenience
- mimics `announce` logic in image callback since we want it there too, but doesn't do validations
- once #259 is completed, we can come back to the workers' announcements & make 'em better
- Original ticket mentions Feeder keeping the last valid state for stories, so presumably once this is finished up we'll need to open a ticket in Feeder for that. 
- Dependent on a PR to add migration to v3, forthcoming!


app/models/audio_file_template.rb
```diff
@@ -20,4 +23,32 @@ def set_defaults
...
+  def validate_audio_file(file)
+    [audio_length_errors(file), label_mismatch_errors(file)].join(' ').strip
+  end
+
+  private
+
+  def audio_length_errors(file)
+    length_errors = ''
+    shorter_than_min = file.length < length_minimum
+    longer_than_max = length_maximum != 0 && file.length > length_maximum
```
cavis: Is there any chance `file.length` is nil?

app/models/audio_file.rb
```diff
@@ -48,4 +35,29 @@ class AudioFile < BaseModel
...
+
+  def compliant_with_template?
+    audio_errors.nil?
+  end
+
+  def validate_on_template
+    # don't bother checking template for audio file that's not yet ready
+    # note: this does mean that files previously deemed invalid by template
+    # will never make it past this line & will stay INVALID.
+    return if [NOTFOUND, VALIDATING, INVALID, FAILED].include?(status)
```
cavis: Is there a reason not to re-run the checks when `INVALID`?  Just to save some time?

app/models/audio_file_template.rb
```diff
@@ -20,4 +23,32 @@ def set_defaults
...
+
+    if shorter_than_min || longer_than_max
+      length_errors << "Audio file #{file.position} '#{file.label}' is " +
```
cavis: Maybe a bit verbose, with the label in there multiple times.  And is the position necessary?  (Since the labels _should_ match between the template and the file).

app/models/audio_file_template.rb
```diff
@@ -20,4 +23,32 @@ def set_defaults
...
+                       "#{file.length.to_time_string}, but the '#{file.label}' " +
+                       "must be between #{length_minimum.to_time_string} " +
+                       "and #{length_maximum.to_time_string}."
+    end
+    length_errors
+  end
+
+  def label_mismatch_errors(file)
```
cavis: In what scenario would this error be important?

If you re-labeled one of your audio-file-templates, this would cause Feeder to unpublish every episode of your podcast.  I was thinking we would only unpublish them if you re-durationed one of the templates, and the file didn't meet the new requirement.

app/models/audio_version.rb
```diff
@@ -20,4 +25,33 @@ def length(reload=false)
...
+    file_errors.nil?
+  end
+
+  private
+
+  def validate_on_template
+    return unless audio_version_template
+
+    noncompliant_file = audio_files.find do |af|
+      !af.compliant_with_template? || af.status == INVALID
```
cavis: Won't an `INVALID` file always be non-compliant?  Or am I missing something?

Also ... does this need to account for the other incomplete audio statuses?  (NOTFOUND, VALIDATING, FAILED).  Wondering if we need to propagate those statuses up as well.

app/models/audio_version.rb
```diff
@@ -20,4 +25,33 @@ def length(reload=false)
...
+    end
+    if noncompliant_file
+      self.status = INVALID
+      self.file_errors = noncompliant_file.audio_errors ||
```
cavis: Alternately, could comma-join every file error here, instead of finding the first.

app/models/audio_version_template.rb
```diff
@@ -26,8 +31,65 @@ def self.policy_class
     SeriesAttributePolicy
   end
 
+  def set_defaults
+    self.length_minimum ||= 0
```
cavis: I guess i thought this was already happening.  Good catch!

app/models/audio_version_template.rb
```diff
@@ -26,8 +31,65 @@ def self.policy_class
...
+    if shorter_than_min || longer_than_max
+      length_errors << "Duration of audio version '#{audio_version.label}' is " +
+                       "#{audio_version.length.to_time_string}, but the " +
+                       "'#{audio_version.label}' must be between " +
+                       "#{length_minimum.to_time_string} and #{length_maximum.to_time_string}."
+    end
+    length_errors
+  end
+
+  def label_mismatch_errors(audio_version)
```
cavis: Again, not entirely sure label changes should invalidate everything

app/models/concerns/validity_flag.rb
```diff
@@ -0,0 +1,26 @@
+# encoding: utf-8
+
+module ValidityFlag
```
cavis: 👍

app/representers/api/story_representer.rb
```diff
@@ -15,6 +15,8 @@ class Api::StoryRepresenter < Api::BaseRepresenter
   property :duration, writeable: false
   property :points, writeable: false
   property :app_version, writeable: false
+  property :status, writeable: false
+  property :version_errors, writeable: false
```
cavis: Should we also represent the error messages for files/versions?

db/schema.rb
```diff
@@ -124,6 +124,7 @@
     t.string   "channel_mode",     limit: 255
     t.string   "status",           limit: 255
     t.string   "format",           limit: 255
+    t.text     "audio_errors",     limit: 65535
```
cavis: Wondering if it would be worth standardizing this, since it's always the human-readable version of the `status` field.  Maybe something like `status` and `status_msg`?  Could let us store other human readable things in there down the line, to indicate why `status` is what it is.

app/models/audio_file_template.rb
gcampo88: Not sure how common it'll be, but I'll account for it in case!

app/models/audio_file_template.rb
gcampo88: 👍  Will remove position and only mention label once

app/models/audio_file_template.rb
gcampo88: Good call. Will nix!

app/models/audio_file.rb
gcampo88: Yeah, just to save some time. But I can remove this line! Now that we're not checking labels any more, and are checking for a nil file.length, not as worried about nil file errors.

app/models/audio_version.rb
gcampo88: `compliant_with_template` currently checks only the `audio_errors`. So for now, this line is looking for either a file that's uploaded successfully but doesn't comply with the template, or a file has uploaded unsuccessfully & been deemed invalid by the lambda callback. 

as for whether it should account for notfound/ validating/ failed -- could make sense to do so! what do you think?

app/models/audio_version.rb
gcampo88: 👍

app/models/audio_version_template.rb
gcampo88: good call

db/schema.rb
gcampo88: 👍

app/representers/api/story_representer.rb
gcampo88: :+1:

gcampo88:
back to you @cavis: 
-made stories, files, and versions all have `status` and `status_msg` (haven't hooked the lambda callback worker into using the `status_msg` yet)
-stopped doing any validation based on labels
-other various small fixes

app/models/audio_file.rb
```diff
@@ -48,4 +35,24 @@ class AudioFile < BaseModel
   def fixerable_final?
     [VALID, COMPLETE, TRANSFORMING, TRANSFORM_FAILED, TRANSFORMED].include? status
   end
+
+  def compliant_with_template?
+    status_msg.nil?
+  end
+
+  def validate_on_template
```
cavis: For some reason, this method is still bugging me.  If I just created a new audio_file ... the status is going to be UPLOADED, while we're waiting for the worker.  Will this change it to INVALID?  Will this nil out a status_msg that the lambda may have set?

@kookster you have any thoughts on this?

I _think_ the current possible state transitions (saving with a new file, background working, then validating) should look like this:

1. UPLOADED -> COMPLETE (lambda worked, template validated)
2. UPLOADED -> NOTFOUND (lambda couldn't download file)
3. UPLOADED -> FAILED (lambda couldn't process file)
4. UPLOADED -> INVALID (lambda processed file, but decided it was invalid for some reason)
5. UPLOADED -> INVALID (lambda worked, but template invalidated)

I almost think we need to nuke case (4), so we're not reusing the INVALID to mean 2 different things.

Not entirely sure how to address the overwriting-statuses/messages problem.  Anyone else have an idea?

gcampo88:
@cavis I think the idea of explicitly calling this `validate_on_template` method only from the audio callback worker could work! It would mean that if a template changed, the validation on it would not run again -- but maybe that's what we want? To allow formerly compliant audio files to remain as they were...?

yeah, open to more ideas on how / where to do this.

app/workers/audio_callback_worker.rb
```diff
@@ -41,6 +42,9 @@ def perform(_sqs_msg, job)
 
     Shoryuken.logger.info("Updating #{job['type']}[#{audio.id}]: status => #{audio.status}")
     audio.save!
+    # announce the audio changes using the story as resource.
+    announce(:audio, :update, id: audio.story.id, resource: audio.story)
```
gcampo88: Is this how we're envisioning this? Or should the subject of the announce be the story?

app/models/audio_file.rb
gcampo88: I've nuked case #4 now, so INVALID will only mean invalid by template. if the lambda has determined the file to be invalid, it'll be called FAILED.

app/workers/audio_callback_worker.rb
gcampo88: Actually -- will having the subject here be `:audio` end up trapping us in an infinite loop of the audio announcement triggering the lambda triggering the audio announcement etc?

gcampo88:
back to you @cavis and/or @kookster, with a question about how I've set up the announcements in the callbacks (should subject be audio or story/parent?)

app/workers/audio_callback_worker.rb
cavis: Now that I think about it, it wouldn't actually create an infinite loop, since the lambda ignores any announcements where `status != UPLOADED`.

But ... I don't think we actually need/should announce the audio here.  I think we just want to `announce(:story` so that feeder gets notified that this file finished processing.

cavis: So this would fire off an resource=story announcement on the `audio` channel?

Yeah, I _think_ it should be `announce(:story, :update, id: story.id, audio.story)`.

app/workers/image_callback_worker.rb
```diff
@@ -17,18 +19,21 @@ def perform(_sqs_msg, job)
...
     else
       image.upload_path = nil
-      image.status = Image::COMPLETE
+      image.status = COMPLETE
     end
 
     Shoryuken.logger.info("Updating #{job['type']}[#{image.id}]: status => #{image.status}")
     image.save!
+    # announce the audio changes using the story as resource -- or do we want series?
+    announce(:image, :update, id: image.story.id, resource: image.story)
```
cavis: But images don't have a story, right?  Or at least not all of them do?  (This callback is shared by series_images, story_images, account_images, etc).

This should announce on the parent channel ... but we probably only care about series and story here.  So maybe detect if the image is one of those, and fire off the announcement for just those?

test/workers/audio_callback_worker_test.rb
```diff
@@ -79,20 +87,31 @@ def perform(attrs = {})
...
     perform(processed: false)
-    audio.status.must_equal AudioFile::FAILED
+    audio.status.must_equal AudioCallbackWorker::FAILED
     audio.fixerable_final?.must_equal false
   end
 
+  it 'announces audio updates with story as resource' do
+    perform(name: 'foo.bar')
+    last_message.wont_be_nil
+    last_message['subject'].must_equal :audio
```
cavis: Yeah, i think this should be `:story`, right?

test/workers/image_callback_worker_test.rb
```diff
@@ -56,20 +60,29 @@ def perform(attrs = {})
...
     perform(resized: false)
-    image.status.must_equal Image::FAILED
+    image.status.must_equal ImageCallbackWorker::FAILED
     image.fixerable_final?.must_equal false
   end
 
+  it 'announces image updates with story as resource' do
+    perform(name: 'foo.bar')
+    last_message.wont_be_nil
+    last_message['subject'].must_equal :image
```
cavis: Again here...

app/workers/image_callback_worker.rb
gcampo88: [I am ashamed!](https://www.youtube.com/watch?v=oVW4fDOgah4)

test/workers/audio_callback_worker_test.rb
gcampo88: that makes sense to me.

cavis:
+1 looks good to me!

PR #257, opened 2017-02-22
make V4 series deleted, and so hidden from V3

kookster:


cavis:
Whoops - didn't actually mean to request changes there!

kookster:
I think I didn't bother earlier b/c I didn't see any or much harm to having the series appear in V3 - empty series don't show up in search and are basically invisible to listeners.

But it could be super confusing if the same account and a similar series is used for publishing V3 stories.

PR #261, opened 2017-02-27
Make "complete" the end status for stories + audio

### LINKED ISSUE #260: Make "complete" the end status for stories + audio
audio files should be: 
UPLOADED -> MP3S_CREATED -> COMPLETE
instead of UPLOADED -> COMPLETE -> VALID as they are now.

audio versions, stories should also be COMPLETE not VALID when they're ready.

gcampo88:
fixes #260


kookster:
yes, better! I say merge

PR #263, opened 2017-02-28
Announcement fixes

### LINKED ISSUE #258: Allow announcing of same action on multiple resources
```Ruby
    it 'prevents dupe announcements' do
      controller_class.announce_actions(:destroy)
      controller_class.announce_actions(:destroy, :update)
      controller_class.announced_actions.size.must_equal 2
    end

```
^^ no longer want this, since we want to be able to do things like announce an image update on both the image and its parent series. Needed for #243 and #244.

gcampo88:
fixes #258 and  #243


app/controllers/api/series_images_controller.rb
```diff
@@ -6,6 +6,11 @@ class Api::SeriesImagesController < Api::BaseController
   filter_resources_by :series_id
 
   announce_actions decorator: Api::Msg::ImageRepresenter, subject: :image
+  announce_actions :update, resource: :series_resource
```
cavis: Wondering if we should add these same parent-announcements to `story_images` and `audio_files`.

But now that I think about it... Feeder is going to ignore this initial `update` announcement anyways, since the `image.status !== COMPLETE`.  (Or at least it should).

I think we still need this functionality ... just don't need to add it for stories yet.

gcampo88: currently both the Image Callback Worker and the Audio Callback Worker manually announce their parent at the end of the `perform`

PR #267, opened 2017-03-02
Import fixes

kookster:
add user to podcast import
set user and account on story
fire off announce messages

depends on https://github.com/PRX/prx.org/pull/293

### DIFF
--- a/app/controllers/api/auth/podcast_imports_controller.rb
+++ b/app/controllers/api/auth/podcast_imports_controller.rb
@@ -9,6 +9,12 @@ class Api::Auth::PodcastImportsController < Api::BaseController
 
   filter_params :url
 
+  def create_resource
+    super.tap do |podcast_import|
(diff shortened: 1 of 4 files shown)


app/controllers/api/auth/podcast_imports_controller.rb
```diff
@@ -9,6 +9,12 @@ class Api::Auth::PodcastImportsController < Api::BaseController
 
   filter_params :url
 
+  def create_resource
+    super.tap do |podcast_import|
+      podcast_import.user_id ||= current_user.id
```
cavis: There's a user_id on podcast imports?  I thought it was just account and series.

cavis: Ah, there it is... schema.rb!

kookster: Yes, this PR is what adds it, needed to set story.creator_id

PR #270, opened 2017-03-03
Call save on parent objects instead of relying on touch to do it

gcampo88:
Should fix the bug in import where the story status is "invalid" despite all children being "complete"


gcampo88:
@kookster any reason to not go for the quick fix on this?

kookster:
#269 has an alternative solution

gcampo88:
yup, that's better. closing this!

PR #280, opened 2017-03-07
Import better info

### LINKED ISSUE #275: On import, populate feedburner link as the public url
If the import, if we are importing a feedburner url, save the feedburner url as the public url link

Question - will this mess up the player or anything if this is not actually yet the public url, since on import the feedburner url is still pulling from some other feed, not this new one? @cavis @gcampo88

### LINKED ISSUE #277: On import, set feeder owner and author name and email
currently it is using the itunes:author tag, which has no email
look at the author tag as well

### LINKED ISSUE #278: On import, use description or itunes:summary for podcast/episode description
In some podcasts, the description is the same as summary and in some it is the same as subtitle.
Instead of looking at description first, see if there is an itunes summary, and if it is larger than description, and use that instead of the description to set the cms resource description.

gcampo88:
fixes #275, fixes #278, fixes #277

### DIFF
--- a/app/models/podcast_import.rb
+++ b/app/models/podcast_import.rb
@@ -127,7 +127,7 @@ class PodcastImport < BaseModel
     podcast_attributes[:new_feed_url] = feed.itunes_new_feed_url
     podcast_attributes[:path] ||= feed.feedburner_name
 
-    podcast_attributes[:author] = person(feed.itunes_author)
+    podcast_attributes[:author] = feed.author ? person(feed.author) : person(feed.itunes_author)
     podcast_attributes[:managing_editor] = person(feed.managing_editor)
     podcast_attributes[:owners] = Array(feed.itunes_owners).map do |o|
       { name: o.name, email: o.email }
--- a/app/models/podcast_import.rb
+++ b/app/models/podcast_import.rb
@@ -67,12 +67,14 @@ class PodcastImport < BaseModel
   end
 
   def create_series_from_podcast(feed = self.feed)
+    description = feed.itunes_summary ? [feed.description, feed.itunes_summary].max : feed.description
+
     # create the series
     self.series = create_series!(
       account: account,
       title: feed.title,
       short_description: feed.itunes_subtitle,
-      description: feed.description
+      description: description
     )
     save!
 
--- a/app/models/podcast_import.rb
+++ b/app/models/podcast_import.rb
@@ -128,6 +128,8 @@ class PodcastImport < BaseModel
     podcast_attributes[:explicit] = feed.itunes_explicit
     podcast_attributes[:new_feed_url] = feed.itunes_new_feed_url
     podcast_attributes[:path] ||= feed.feedburner_name
+    podcast_attributes[:feedburner_url] ||= feed.feedburner_url
+    podcast_attributes[:url] ||= feed.feedburner_url
 
     podcast_attributes[:author] = feed.author ? person(feed.author) : person(feed.itunes_author)
     podcast_attributes[:managing_editor] = person(feed.managing_editor)
--- a/app/models/podcast_import.rb
+++ b/app/models/podcast_import.rb
@@ -67,14 +67,14 @@ class PodcastImport < BaseModel
   end
 
   def create_series_from_podcast(feed = self.feed)
-    description = feed.itunes_summary ? [feed.description, feed.itunes_summary].max : feed.description
+    description = [feed.description, feed.itunes_summary].max if feed.itunes_summary
 
     # create the series
     self.series = create_series!(
       account: account,
       title: feed.title,
       short_description: feed.itunes_subtitle,
-      description: description
+      description: description || feed.description
     )
     save!
 
--- a/app/models/podcast_import.rb
+++ b/app/models/podcast_import.rb
@@ -132,12 +132,12 @@ class PodcastImport < BaseModel
     podcast_attributes[:explicit] = feed.itunes_explicit
     podcast_attributes[:new_feed_url] = feed.itunes_new_feed_url
     podcast_attributes[:path] ||= feed.feedburner_name
-    podcast_attributes[:feedburner_url] ||= feed.feedburner_url
-    podcast_attributes[:url] ||= feed.feedburner_url
+    podcast_attributes[:feedburner_url] ||= feedburner_url(feed.feedburner_name)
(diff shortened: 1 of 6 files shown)


kookster:
I have 2nd thoughts about PRX/cms.prx.org#277

kookster:
@gcampo88 please update for new understanding of #277 - oh the things we learn!

app/models/podcast_import.rb
```diff
@@ -126,12 +128,12 @@ def create_podcast
...
+    podcast_attributes[:feedburner_url] ||= feed.feedburner_url
+    podcast_attributes[:url] ||= feed.feedburner_url
 
-    podcast_attributes[:author] = person(feed.itunes_author)
+    podcast_attributes[:author] = feed.author ? person(feed.author) : person(feed.itunes_author)
     podcast_attributes[:managing_editor] = person(feed.managing_editor)
-    podcast_attributes[:owners] = Array(feed.itunes_owners).map do |o|
-      { name: o.name, email: o.email }
-    end
+    podcast_attributes[:owner] = person(feed.owner)
```
gcampo88: note this `person()` call is redundant since feed.owner will already be in the { name: name, email: email } format -- I left it in there for symmetry with the author and managing editor attribute settings, so someone glancing just at this will know that those 3 will all have the same structure.

gcampo88:
alrighty @kookster, #277 rehandled!

app/models/podcast_import.rb
```diff
@@ -67,12 +67,14 @@ def validate_feed(podcast_feed)
...
   def create_series_from_podcast(feed = self.feed)
+    description = [feed.description, feed.itunes_summary].max if feed.itunes_summary
+
     # create the series
     self.series = create_series!(
       account: account,
       title: feed.title,
       short_description: feed.itunes_subtitle,
-      description: feed.description
+      description: description || feed.description
```
kookster: seems odd to set description based on `feed.description`, but then to also have it in this expression.

app/models/podcast_import.rb
```diff
@@ -67,12 +67,14 @@ def validate_feed(podcast_feed)
   end
 
   def create_series_from_podcast(feed = self.feed)
+    description = [feed.description, feed.itunes_summary].max if feed.itunes_summary
```
kookster: This `max` is checking alphabetical order, not string length
```
irb(main):003:0> ['abcdefg', 'def'].max
=> "def"
```

app/models/podcast_import.rb
kookster: +1

lib/feedjira/parser/podcast.rb
```diff
@@ -56,6 +57,10 @@ class Podcast
       element :"media:keywords", as: :media_keywords
       elements :"media:category", as: :media_categories
 
+      def feedburner_url
```
kookster: I don't think this goes in the parser

lib/feedjira/parser/podcast.rb
```diff
@@ -37,7 +38,7 @@ class Podcast
       element :"itunes:explicit", as: :itunes_explicit
       element :"itunes:complete", as: :itunes_complete
       element :"itunes:new_feed_url", as: :itunes_new_feed_url
-      elements :"itunes:owner", as: :itunes_owners, class: ITunesRSSOwner
+      element :"itunes:owner", as: :itunes_owner, class: ITunesRSSOwner
```
kookster: maybe leave this as multiple
https://github.com/feedjira/feedjira/blob/master/lib/feedjira/parser/itunes_rss.rb#L60

test/models/podcast_import_test.rb
```diff
@@ -58,7 +60,7 @@
     importer.distribution.audio_version_template.must_equal template
   end
 
-  it 'creates a podast' do
+  it 'creates a podcast' do
```
kookster: I can't spel!

lib/feedjira/parser/podcast.rb
```diff
@@ -74,6 +79,10 @@ def parse_datetime(value)
         nil
       end
 
+      def owner
```
kookster: won't need this either if `owners` retained

lib/feedjira/parser/podcast.rb
```diff
@@ -19,6 +19,7 @@ class Podcast
       element :description
       element :language
       element :copyright
+      element :author
```
kookster: `author` is not at channel level, or at least it shouldn't be.
http://cyber.harvard.edu/rss/rss.html

app/models/podcast_import.rb
```diff
@@ -126,12 +128,12 @@ def create_podcast
     podcast_attributes[:explicit] = feed.itunes_explicit
     podcast_attributes[:new_feed_url] = feed.itunes_new_feed_url
     podcast_attributes[:path] ||= feed.feedburner_name
+    podcast_attributes[:feedburner_url] ||= feed.feedburner_url
+    podcast_attributes[:url] ||= feed.feedburner_url
 
-    podcast_attributes[:author] = person(feed.itunes_author)
+    podcast_attributes[:author] = feed.author ? person(feed.author) : person(feed.itunes_author)
```
kookster: I know we talked about this, and I probably said this wrong, but there isn't a `feed.author`

app/models/podcast_import.rb
gcampo88: huh, so it is!

app/models/podcast_import.rb
gcampo88: 👍

lib/feedjira/parser/podcast.rb
gcampo88: ah, gotcha. OK, so import currently can't set author_email if it's only using `<itunes:author>`.

lib/feedjira/parser/podcast.rb
gcampo88: will move to PodcastImport

lib/feedjira/parser/podcast.rb
kookster: nope, but so long as it sets owner:email after your fix, we'll have that correctly

lib/feedjira/parser/podcast.rb
gcampo88: 👍

gcampo88:
ok @kookster, think I got 'em all! back to you.

PR #281, opened 2017-03-08
Send s3 audio urls in announce message for feeder

kookster:
This will allow feeder to use these s3://bucket/path/file.ext style urls for sending to fixer.

### DIFF
--- a/app/controllers/concerns/announce_actions/announce_filter.rb
+++ b/app/controllers/concerns/announce_actions/announce_filter.rb
@@ -65,6 +65,7 @@ module AnnounceActions
 
     def decorator_class(name)
       "Api::Msg::#{name}Representer".safe_constantize ||
+        "Api::Auth::#{name}Representer".safe_constantize ||
         "Api::#{name}Representer".safe_constantize ||
         "Api::Min::#{name}Representer".safe_constantize
     end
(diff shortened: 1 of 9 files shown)


app/controllers/concerns/announce_actions/announce_filter.rb
```diff
@@ -65,6 +65,7 @@ def decorator_for_model(model)
 
     def decorator_class(name)
       "Api::Msg::#{name}Representer".safe_constantize ||
+        "Api::Auth::#{name}Representer".safe_constantize ||
```
cavis: Just to sanity check - seem to be no implications for existing announcements here.  Looks like `Auth::Series` and `Auth::Story` just rewrite some links, which won't impact anything.

app/representers/api/auth/story_representer.rb
```diff
@@ -18,4 +18,17 @@ def self_url(r)
...
     } if !represented.published_at.nil?
   end
+
+  link :audio do
+    {
+      href: api_authorization_audio_files_path(represented.id),
+      count: represented.default_audio.count
+    } if represented.id
+  end
+  embed :default_audio,
```
cavis: A bit off-topic... but are we assuming that `default_audio` will always be the podcast version?

I can imagine at some point we'll want a link to auth-story-versions in here.  A task for another day!

app/controllers/concerns/announce_actions/announce_filter.rb
kookster: yeah, if anything would help as wouldn't need to get auth version some special way.
I also looked through, there isn't an auth audio_file or image representer, so shouldn't break lambdas either.

app/representers/api/auth/story_representer.rb
kookster: agreed, this is not for always - this will have to be more clever at some point!

PR #289, opened 2017-03-10
Announce story and series with public self uri

### LINKED ISSUE #288: Announce must use representer public self link href
Using the auth representers prevents feeder from processing the changes, as the self link with `/authorization` in it doesn't match the `prx_uri` saved for the podcast or episode in feeder.

kookster:
fixes #288 
- [x] Make new msg representers for series and story
- [x] On both, send the public self link
- [x] Don't send the stories embedded in the series - long & redundant with story announcements
- [x] Revert announce filter, no longer try to use auth representers
- [x] Update jobs to use msg not auth representers

### DIFF
--- a/app/representers/api/msg/series_representer.rb
+++ b/app/representers/api/msg/series_representer.rb
@@ -1,0 +1,11 @@
+# encoding: utf-8
+
+class Api::Msg::SeriesRepresenter < Api::SeriesRepresenter
+  # don't embed the stories, makes this a much smaller message
+  embed :public_stories,
+        as: :stories,
+        paged: true,
+        item_class: Story,
+        item_decorator: Api::Min::StoryRepresenter,
+        zoom: false
+end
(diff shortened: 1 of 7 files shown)


app/representers/api/msg/series_representer.rb
```diff
@@ -0,0 +1,11 @@
+# encoding: utf-8
+
+class Api::Msg::SeriesRepresenter < Api::SeriesRepresenter
+  # don't embed the stories, makes this a much smaller message
+  embed :public_stories,
+        as: :stories,
+        paged: true,
+        item_class: Story,
+        item_decorator: Api::Min::StoryRepresenter,
+        zoom: false
```
gcampo88: so it's the `zoom: false` that dictates stories won't be embedded by default, right? and then if the representer is passed the zoom param, then it'll embed stories?

kookster: that's the idea, but since this is used in the announcement, there is no param passed in, so effectively this will mean it won't ever be embedded.

Issue #299, opened 2017-03-13
Send default episode URL to feeder


cavis:
On a `POST /stories` request ...

When creating the episode in feeder, also send a default `episode.url` link.  Should be the story-alt link, to the location in beta.

Only do this on controller POST requests.  Not on the background import.

cavis:
Goal here is to get a default `beta.prx.org` link in the "Episode URL" field in Publish.  We can just always send this field when POST-ing to Feeder, since Publish's subsequent PUT to Feeder (to update any fields you set) will overwrite.

cavis:
TODO: verify in prod after deploy

cavis:
Confirmed in prod:

- [x] New episodes with a blank Episode Url get defaulted to beta
- [x] If you set the Url, the subsequent PUT to feeder overwrites the beta default

PR #303, opened 2017-03-17
Add default URL to episodes on feeder POST reqs, but not import

gcampo88:
- #299 adds a default URL to episodes POST request (but not on podcast import) if ENV['PRX_HOST'] is set.

### DIFF
--- a/app/models/story_distributions/episode_distribution.rb
+++ b/app/models/story_distributions/episode_distribution.rb
@@ -49,6 +49,6 @@ class StoryDistributions::EpisodeDistribution < StoryDistribution
 
   def default_url(story)
     path = "#{story.class.name.underscore.pluralize}/#{story.id}"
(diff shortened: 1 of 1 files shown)


app/models/story_distributions/episode_distribution.rb
```diff
@@ -13,7 +13,9 @@ def distribute!
   def add_episode_to_feeder(create_attributes = {})
     return unless url.blank?
     podcast = distribution.get_podcast
-    episode = podcast.episodes.post(episode_attributes.merge(create_attributes))
+    post_attributes = episode_attributes.merge(create_attributes)
```
cavis: Is there a reason not to just put a default `url` in the `episode_attributes`?

It looks like the PodcastImport is always going to set `create_attributes[:url]` to at least nil.  So it should always override the beta-default in the merge call then, right?  (`{k: 'val'}.merge({k: nil}) == {k: nil}`)

PR #309, opened 2017-04-18
Replace audio on podcast import

kookster:
- [x] Add `config` for a `PodcastImport`, use for # of segments and replacement audio
- [x] Use this config to setup audio template, and set audio files, when available

### DIFF
--- a/app/models/podcast_import.rb
+++ b/app/models/podcast_import.rb
@@ -118,7 +118,7 @@ class PodcastImport < BaseModel
       length_maximum: 0
     )
 
-    num_segments = [config[:segments].to_i, 1 ].max
+    num_segments = [config[:segments].to_i, 1].max
     num_segments.times do |x|
       num = x + 1
       template.audio_file_templates.create!(
@@ -382,8 +382,8 @@ class PodcastImport < BaseModel
     @uri ||= Addressable::URI.parse(url)
   end
 
-  def connection(uri = uri)
-    conn_uri = "#{uri.scheme}://#{uri.host}:#{uri.port}"
+  def connection(u = self.uri)
+    conn_uri = "#{u.scheme}://#{u.host}:#{u.port}"
     Faraday.new(conn_uri) { |stack| stack.adapter :excon }.tap do |c|
       c.headers[:user_agent] = 'PRX CMS FeedValidator'
     end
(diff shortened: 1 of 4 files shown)


kookster:
Here is the import file for TWWW
http://andrew-s3.prx.org.s3.amazonaws.com/twww/twww_config.json

kookster:
related https://github.com/PRX/prx.org/pull/298

app/models/podcast_import.rb
```diff
@@ -262,9 +280,15 @@ def create_story(entry, series)
       explicit: explicit(entry[:itunes_explicit])
     )
 
-    # add the audio
-    if enclosure = enclosure_url(entry)
-      audio = version.audio_files.create!(label: 'Segment A', upload: enclosure)
+    if config[:audio] && config[:audio][entry.entry_id]
```
cavis: So just to confirm ... if an entry_id is missing from the config, you're going to silently get a single-segment audio version.  (Breaking the validation rules).

Or does this actually throw a "not enough segments" validation error against the audio version template?

test/models/podcast_import_test.rb
```diff
@@ -71,6 +78,7 @@
   end
 
   it 'creates stories' do
+    importer.set_config_url('http://test.prx.org/transistor_import_config.json')
```
cavis: Is this necessary for this test?  Wouldn't you get the exact same output if you didn't set a config?

app/models/podcast_import.rb
kookster: that's right, this isn't requiring the config include entries for every episode.
e.g. you may only be replacing a single segment with another, clean/edited segment, and not have to do that for every one.

I don't think it throws errors, it would set a status and message on the version.

test/models/podcast_import_test.rb
kookster: the config is used in that test, I'll add proof!

kookster: latest commit tests audio url to be same as the one in the config

app/models/podcast_import.rb
cavis: Works for me.  I can see how that could be useful to replace a single segment.

test/models/podcast_import_test.rb
cavis: Prove'd!

PR #311, opened 2017-04-20
Include audio versions in story json

kookster:
Need the audio version info for use in setting the podcast episode `explicit`


kookster:
@gcampo88 updated, figured out why the test was breaking the representers in the audio callback job

PR #320, opened 2017-06-22
Fix announcing story destroy events

kookster:
- [x] Change `delete` to `destroy` in announce actions config
- [x] Also destroy dependencies for the story
- [x] Do not try to retrieve relations when getting the url for a nested (e.g. story) image, they may be deleted
- [x] Fix bug where `producer` relation on story had wrong foreign key

### DIFF
--- a/app/models/story.rb
+++ b/app/models/story.rb
@@ -40,17 +40,17 @@ class Story < BaseModel
   belongs_to :network
 
   has_many :images,
-    -> { where(parent_id: nil).order(:position) },
-    class_name: 'StoryImage',
-    foreign_key: :piece_id,
-    dependent: :destroy
+           -> { where(parent_id: nil).order(:position) },
+           class_name: 'StoryImage',
+           foreign_key: :piece_id,
+           dependent: :destroy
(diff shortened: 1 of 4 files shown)


app/models/story.rb
```diff
@@ -39,23 +39,36 @@ def self.paranoia_scope
...
-  has_many :picks, foreign_key: 'playlistable_id'
+  has_many :picks, foreign_key: 'playlistable_id', dependent: :destroy
   has_many :playlists, through: :picks
   has_many :purchases, foreign_key: :purchased_id
-  has_many :distributions, class_name: StoryDistribution, foreign_key: :piece_id
-
-  has_one :promos, -> { where(promos: true) }, class_name: 'AudioVersion', foreign_key: :piece_id
-  has_one :license, foreign_key: :piece_id
+  has_many :distributions,
+           class_name: StoryDistribution,
```
gcampo88: would vote we make this a string for consistency with the rest of the repo (plus [this](https://stackoverflow.com/questions/20390991/activerecord-pass-class-instead-of-string-to-class-name-when-defining-associati) says so) 
`class_name: 'StoryDistribution',`

(officially nitpickier than hound! 🐶 )

PR #321, opened 2017-06-22
override self_url to include ids of deleted resources

kookster:


kookster:
@gcampo88 this should hopefully fix the latest bug making deletes not work, hopefully the last bug for that problem...

kookster:
I added it to the 2 that needed it.

- Andrew

PR #326, opened 2017-07-26
Check audio files match on channel, bit rate, freq

gcampo88:
Dovetail needs audio files to match on channels, bitrate, and frequency to stitch them together without possibility of error in playback. This checks to make sure all audio files for a given audio version match each other on those three attributes. (first half of https://github.com/PRX/publish.prx.org/issues/432)

The chain of methods to make this happen is:
- audio file gets uploaded, with `status: 'uploaded'`
- audio lambda kicks in and determines length, size, bit rate, frequency, channel etc 
- AudioCallbackWorker adds those attributes to the audio file, sets the file's status to `'mp3s created'` (aka `TRANSFORMED`) and calls `save!`
- before_save hook calls `set_status` on the file and validates it against its own audio file template, and sets the status to `complete` if no error
- after_save hook calls `update_version_status` since the status of the audio file has changed, which calls `save!` on the audio version the file belongs to 
- before_save hook on the audio version calls `set_status`, which will now check for file format mismatches in addition to validating against template

I'm thinking that we can rely on this chain of callbacks to work as expected -- are there any times that an audio file's status would _not_ change, but we'd still fear a new file format mismatch?

### DIFF
--- a/app/models/audio_version.rb
+++ b/app/models/audio_version.rb
@@ -35,7 +35,7 @@ class AudioVersion < BaseModel
   private
 
   def audio_formats_match?
-    [:frequency, :bit_rate, :channel_mode].each do |format|
+    %i(frequency bit_rate channel_mode).each do |format|
       return false if audio_files.map(&format).compact.uniq.length > 1
     end
     true
--- a/app/models/audio_version.rb
+++ b/app/models/audio_version.rb
@@ -34,6 +34,13 @@ class AudioVersion < BaseModel
 
   private
 
+  def audio_formats_match?
+    [:frequency, :bit_rate, :channel_mode].each do |format|
+      return false if audio_files.map(&format).compact.uniq.length > 1
+    end
+    true
+  end
+
   def update_story_status
     story.try(:save!)
   end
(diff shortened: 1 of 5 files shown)


app/models/audio_version.rb
```diff
@@ -48,6 +55,12 @@ def set_status
       return
     end
 
+    if !audio_formats_match?
+      self.status = INVALID
+      self.status_message = 'Mismatch in audio formats between audio files'
```
cavis: Might be nice to indicate which thing(s) didn't match - or just the first thing.  "Audio format" is a bit ambiguous... I might confuse that for meaning "file 1 is mp3 and file 2 is wav".

Guessing we'll update the UI to display the freq/rate/channels for each file.  And maybe even highlight the mismatch in red.  So maybe I'm overthinking this part.

gcampo88:
update -- per tech-cms slack channel on 7/26, looks like variable bitrate may be okay. and: will want to handle when the mismatching file is the zero length silent file that we're adding in import for some shows. the info for that one is: `32 / Stereo / 44.1`. 
will leave this in limbo for now as we wait to decide whether to check bitrate, when / how to show validation errors in publish, and what to do about existing mismatched audio.

cavis:
update 2: even though dovetail _can_ deal with variable bitrates, we shouldn't be allowing it here.

And ... should we also be comparing `content_type` and `layer` on the files?

cavis:
Merging to try this out in staging.  Since prod deploys are still manual, no risk of accidentally doing that.

PR #330, opened 2017-08-31
Create series for feeder podcast

kookster:
- [x] Retrieve feeder podcast
- [x] Create series record
- [x] copy images
- [x] create version template
- [x] create the distribution associated to feeder podcast

### DIFF
--- a/lib/feeder_importer.rb
+++ b/lib/feeder_importer.rb
@@ -3,9 +3,10 @@ require 'addressable/uri'
 
 class FeederImporter
   include PRXAccess
+  include Announce::Publisher
 
   attr_accessor :account_id, :user_id, :feeder_podcast_url
-  attr_accessor :podcast
+  attr_accessor :podcast, :series, :template, :distribution
 
   def initialize(account_id, user_id, feeder_podcast_url)
     self.account_id = account_id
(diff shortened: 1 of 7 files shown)


gcampo88:
@kookster assume I should hold off on review for now? (Github's sending me mixed messages!)

lib/feeder_importer.rb
```diff
@@ -0,0 +1,95 @@
...
+    self.distribution = Distributions::PodcastDistribution.create!(
+      url: feeder_podcast_url,
+      distributable: series,
+      audio_version_template: template
+    )
+
+    series
+  end
+
+  def retrieve_podcast(a_id = account_id, p_url = feeder_podcast_url)
```
gcampo88: any particular reason for making the account id & podcast URL overrideable?

test/lib/feeder_importer_test.rb
```diff
@@ -0,0 +1,69 @@
...
+    stub_import_requests
+  end
+
+
+  it "makes a new importer" do
+    importer.wont_be_nil
+  end
+
+  it "retrieves the feeder podcast" do
+    podcast = importer.retrieve_podcast
```
gcampo88: maybe name this version of `podcast` something slightly different to set it apart from the podcast you declared on line 13?

lib/feeder_importer.rb
gcampo88: & ditto for optional `podcast` parameter for `create_series` def

kookster:
Realized I have tests to fix, assigned too early, sorry!

lib/feeder_importer.rb
kookster: yes, clearer to test if you can call the method without setting the state on the importer

gcampo88: 👍

lib/feeder_importer.rb
```diff
@@ -0,0 +1,95 @@
+require 'prx_access'
+require 'addressable/uri'
+
+class FeederImporter
+  include PRXAccess
+  include Announce::Publisher
+
+  attr_accessor :account_id, :user_id, :feeder_podcast_url
+  attr_accessor :podcast, :series, :template, :distribution
```
gcampo88: assume these two `attr_accessor` calls are kept separate for organization's sake?

kookster: yes, and line length would be > 100 if they were all on one line

gcampo88: 👍

PR #336, opened 2017-09-12
Multiple templates for distributions

kookster:
- [x] Add a content type to the templates
- [x] Add many to many relation between templates
- [x] Allow templates to be set by an array of ids

### DIFF
--- a/app/models/audio_version_template.rb
+++ b/app/models/audio_version_template.rb
@@ -7,6 +7,8 @@ class AudioVersionTemplate < BaseModel
   belongs_to :series, touch: true
 
   has_many :audio_file_templates, -> { order :position }, dependent: :destroy
+  has_many :distribution_templates, dependent: :destroy
+  has_many :distributions, through: :distribution_templates
 
   has_many :audio_versions, dependent: :nullify
   has_many :distributions, dependent: :nullify
--- a/app/models/audio_version_template.rb
+++ b/app/models/audio_version_template.rb
@@ -6,12 +6,10 @@ class AudioVersionTemplate < BaseModel
 
   belongs_to :series, touch: true
 
+  has_many :audio_versions, dependent: :nullify
   has_many :audio_file_templates, -> { order :position }, dependent: :destroy
   has_many :distribution_templates, dependent: :destroy
   has_many :distributions, through: :distribution_templates
-
-  has_many :audio_versions, dependent: :nullify
-  has_many :distributions, dependent: :nullify
 
   after_save :touch_audio_versions
 
--- a/app/models/distribution.rb
+++ b/app/models/distribution.rb
@@ -6,6 +6,8 @@ class Distribution < BaseModel
   belongs_to :audio_version_template
 
   has_many :story_distributions
+  has_many :distribution_templates, dependent: :destroy
+  has_many :audio_version_templates, through: :distribution_templates
 
   serialize :properties, HashSerializer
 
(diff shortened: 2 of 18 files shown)


app/models/distribution_template.rb
```diff
@@ -0,0 +1,14 @@
+# encoding: utf-8
+
+class DistributionTemplate < BaseModel
+  belongs_to :distribution
+  belongs_to :audio_version_template
+
+  def account
+    distribution.try(:account)
```
cavis: Yeah, that makes sense.  Dist-owner is responsible for the relation, rather than the AudioVersion owner.  (Though they're almost certainly the same).

app/representers/api/distribution_representer.rb
```diff
@@ -5,6 +5,18 @@ class Api::DistributionRepresenter < Api::BaseRepresenter
   property :guid
   property :url
   property :kind
+
+  property :set_audio_version_template_uris,
+           readable: false,
+           reader: ->(doc, _args) do
+             distribution_templates.clear
```
cavis: So you're saying the frontend shouldn't just PUT this every time then, right?  Only when it changes.  Since this is going to recreate the join table rows, whether or not it actually changed.

app/models/distribution.rb
```diff
@@ -6,7 +6,8 @@ class Distribution < BaseModel
   belongs_to :audio_version_template
```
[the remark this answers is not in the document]
cavis: Is there a case for switching this to be `audio_version_templates.first` right now?  So we won't need another CMS change to run the schema changes dropping this?

app/representers/api/distribution_representer.rb
cavis: Also wondering what the current set `audio_version_template` will do once the has-one relation disappears.

app/models/distribution.rb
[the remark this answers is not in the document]
kookster: it might work, but it's a relation which has more than a few methods and ways of being accessed. I'll try something naive, but it may not work for anything but read-only and updating the id.

app/models/audio_version_template.rb
```diff
@@ -7,6 +7,8 @@ class AudioVersionTemplate < BaseModel
   belongs_to :series, touch: true
 
   has_many :audio_file_templates, -> { order :position }, dependent: :destroy
+  has_many :distribution_templates, dependent: :destroy
+  has_many :distributions, through: :distribution_templates
```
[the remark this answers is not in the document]
kookster: yeah, meant to remove the dupe.

app/models/distribution_template.rb
kookster: probably need a validation to make sure they both belong to the same series as well

app/representers/api/distribution_representer.rb
cavis: (To be clear, i'm fine with a PUT always replacing them.  Just making sure I know the API!)

kookster: Not sure I can make everything work correctly for both single and multiple relation here.
I was assuming that these change infrequently enough that this is a good strategy.
How after are you saving changes to the distribution?

cavis: For some properties, they're PUT every time you save.  For others (and i _think_ the `set_something_uri` are in this category) they only get serialized when changed.  Will just have to make sure this array-of-uris ends up in that category.

So given the complexity of making both the has-one and has-many work concurrently, it sounds the order of merge/deploy should be:

1. This
2. PRX/publish.prx.org#456 (currently making sure that plays well with these changes)
3. PRX/prx.org#306

Right?

cavis:
:+lgtm!:

PR #339, opened 2017-09-18
Import with different audio templates

kookster:
- [x] Don't create templates up front (ignore `config[:segments]`)
- [x] As a template is needed in story import, find or create it

### DIFF
--- a/app/models/podcast_import.rb
+++ b/app/models/podcast_import.rb
@@ -295,7 +295,7 @@ class PodcastImport < BaseModel
 
       num_segments.times do |x|
         num = x + 1
-        ft = template.audio_file_templates.create!(
+        template.audio_file_templates.create!(
           position: num,
           label: "Segment #{num}",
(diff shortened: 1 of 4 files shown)


app/models/podcast_import.rb
```diff
@@ -301,6 +283,35 @@ def create_story(entry, series)
     story
   end
 
+  def get_or_create_template(segments)
+    num_segments = [segments.to_i, 1].max
+    if !templates[num_segments]
+      template = series.audio_version_templates.create!(
+        label: "Podcast Audio #{num_segments}",
```
gcampo88: Might confuse users to see 'Podcast Audio 3' as their template label if it's their only one -- if it were me, I'd wonder where versions 1 and 2 went. How about `"Podcast Audio #{num_segments} Segments"` ?

kookster: good suggestion, will do.

PR #340, opened 2017-09-19
Import feeder episodes into CMS

kookster:
For shows in crier -> feeder -> dovetail, but not in publish.

Runs in cms, but must be configured to connect to the feeder db to read and write.

Copies files from feeder to cms upload bucket to process as if they are newly uploaded to cms, and so assign the correct filenames, and then copies the file locations in cms back to feeder.


kookster:
ok, @gcampo88 I think hound should be pretty happy now, good to review

lib/feeder_importer.rb
```diff
@@ -71,25 +160,138 @@ def create_series(pcast = podcast)
     end
 
     self.distribution = Distributions::PodcastDistribution.create!(
-      url: feeder_podcast_url,
+      url: podcast_id,
```
gcampo88: why just the id instead of `#{feeder_root}/podcasts/#{podcast_id}`?

lib/feeder_importer.rb
```diff
@@ -71,25 +160,138 @@ def create_series(pcast = podcast)
...
+      account_id: account_id,
+      title: episode.title,
+      short_description: episode.subtitle,
+      description_html: episode.description,
+      tags: episode.categories,
+      published_at: episode.published_at
+    }
+    story = series.stories.create!(attrs)
+
+    version = story.audio_versions.create!(
```
gcampo88: in regular import we apply a label here, right?

lib/feeder_importer.rb
```diff
@@ -1,64 +1,153 @@
...
-  attr_accessor :podcast, :series, :template, :distribution
+  attr_accessor :account_id, :user_id, :podcast_id, :set_episode_urls
+  attr_accessor :podcast, :series, :template, :distribution, :stories
+
+  def debug
+    TRUE
+  end
 
-  def initialize(account_id, user_id, feeder_podcast_url)
+  def initialize(account_id, user_id, podcast_id, set_episode_urls = false)
```
gcampo88: when would set_episode_urls be true?

lib/feeder_importer.rb
```diff
@@ -71,25 +160,138 @@ def create_series(pcast = podcast)
...
+  end
+
+  def audio_file_original_url(af)
+    "s3://#{Rails.env}.mediajoint.prx.org/public/audio_files/#{af.id}/#{File.basename(af.filename)}"
+  end
+
+  def update_podcast
+    podcast.update_attributes(
+      prx_account_uri: "/api/v1/accounts/#{account_id}",
+      prx_uri: "/api/v1/series/#{series.id}",
```
gcampo88: if I'm understanding your comment at top of file correctly, setting the podcast's prx_uri to the link with the newly created series id in it will get us to the point where changing things in CMS will change them in Feeder too, right? In which case, should this be a separate method that we call manually after `import` has completed?

test/lib/feeder_importer_test.rb
```diff
@@ -1,68 +1,51 @@
...
-    remote_podcast = importer.retrieve_podcast(account_id, feeder_podcast_url)
+    remote_podcast = importer.retrieve_podcast
     remote_podcast.wont_be_nil
     remote_podcast.title.must_equal 'Transistor'
   end
 
   it 'creates a series' do
-    series = importer.create_series(podcast)
+    importer.retrieve_podcast
+    podcast = importer.podcast
```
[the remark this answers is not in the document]
gcampo88: add a `podcast.wont_be_nil` to please sir hound?

test/db/feeder_transistor.sql
```diff
@@ -0,0 +1,941 @@
...
+
+
+--
+-- Name: media_resources; Type: TABLE; Schema: public; Owner: feeder
+--
+
+CREATE TABLE media_resources (
+    id integer NOT NULL,
+    episode_id integer,
+    "position" integer,
```
gcampo88: is "position" a reserved word? (why quotes?)

lib/feeder_importer.rb
```diff
@@ -1,64 +1,153 @@
...
-  include PRXAccess
   include Announce::Publisher
+  include PRXAccess
 
-  attr_accessor :account_id, :user_id, :feeder_podcast_url
-  attr_accessor :podcast, :series, :template, :distribution
+  attr_accessor :account_id, :user_id, :podcast_id, :set_episode_urls
+  attr_accessor :podcast, :series, :template, :distribution, :stories
+
+  def debug
```
gcampo88: should nix this before deploy, or do you have testing plans that require it to stay in place?

test/db/feeder_transistor.sql
kookster: idk - this was generated by the `pg_dump` utility, I did not write this.

lib/feeder_importer.rb
kookster: I intentionally am leaving in the debug, but I can set the default to false

lib/feeder_importer.rb
kookster: when we are importing a feed from libsyn, and the episode urls are on libsyn, then we want to set them as libsyn will be retired. When a producer has set the episode urls to go to their own website, and they want to keep those urls, then we would set it to false.

lib/feeder_importer.rb
```diff
@@ -1,64 +1,153 @@
...
       length_maximum: 0
     )
 
-    num_segments = 1
-    episode = pcast.episodes.first
-    if episode.attributes['media']
-      num_segments = [episode.media.count, num_segments].max
+    episode = podcast.episodes.first
+    if episode.media_resources
+      num_segments = [episode.media_resources.count, num_segments].max
```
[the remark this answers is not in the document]
kookster: There should always be 1, as the comment indicates, but I worry that this code might be used for other cases eventually, and wrote the code to handle more if necessary.

Feeder definitely allows more than one file, so I thought it safer to handle what could be the case in feeder, especially since it is a pretty small amount of added code/complexity.

The code to actually add the audio files from feeder to an episode likewise doesn't only get `first`, it looks through all of them, b/c that is what the feeder model supports.

lib/feeder_importer.rb
kookster: yeah, that looks wrong

lib/feeder_importer.rb
```diff
@@ -71,25 +160,138 @@ def create_series(pcast = podcast)
...
     series
   end
 
-  def retrieve_podcast(a_id = account_id, p_url = feeder_podcast_url)
-    self.podcast = api(root: feeder_root, account: a_id).tap { |a| a.href = p_url }.get
+  def create_series_image(podcast_image, purpose)
+    upload_url = copy_image(SecureRandom.uuid, podcast_image)
+    image = series.images.create!(upload: upload_url, purpose: purpose)
+    announce_image(image)
+    podcast_image.update_attribute(:original_url, image.public_url(version: 'original'))
```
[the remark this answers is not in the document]
kookster: yes - there are no validations on any of the feeder models anyway, so `update_attribute` is just fine.

lib/feeder_importer.rb
kookster: sure!

lib/feeder_importer.rb
kookster: That is the effect it will have, but I did not see a compelling reason to wait and add an additional manual step.

The point of import is to get to that point, and I designed this to all work in a single script run rather than needing to come back and fix more things later.

We can be extra cautious and do this manually later, but I didn't want to do that and have any changes in cms discarded by feeder until we set it.

lib/feeder_importer.rb
kookster: yeah, I can look at changing this, but I don't want the tests to actually mess around with files on s3, so debug being true is needed for the tests to run.

lib/feeder_importer.rb
```diff
@@ -71,25 +160,138 @@ def create_series(pcast = podcast)
...
+
+  def copy_file(from_path, to_path)
+    connection = AudioFileUploader.new.send(:storage).connection
+    options = {
+      'x-amz-metadata-directive' => 'COPY',
+      'x-amz-acl' => 'public-read'
+    }
+    if debug
+      # Leaving this - useful for debugging when not wanting to actually update s3
+      # puts "copy_object('prx-feed', #{from_path}', 'prx-up', '#{to_path}', '#{options.inspect}')"
```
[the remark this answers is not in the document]
kookster: it is pretty noisy when the tests run, so I don't want it all the time.

test/lib/feeder_importer_test.rb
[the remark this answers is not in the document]
kookster: hound is wrong or out of date - `podcast` gets used on lines 35 and 40

kookster:
@gcampo88 changed the things

PR #346, opened 2017-09-27
Add season identifiers

kookster:
- [x] Add to the schema (https://github.com/PRX/prx.org/pull/311)
- [x] Add to the story representers
- [x] Send identifiers to feeder from episode distribution
- [x] Strip identifiers to integers to send to feeder


gcampo88:
OK to merge this now since it'll only deploy through to staging @kookster ? or would you rather wait til the prx.org one is merged?

Issue #345, opened 2017-09-26
Audio callback worker for video


cavis:
Remaining audio-callback-worker changes from #306:

- [x] Use the new [callback spec](https://github.com/PRX/cms-audio-lambda#callbacks) for the cms-audio-lambda
- [x] Use the returned `mime` as the content-type instead of CMS detecting it.
- [x] Validate audio-files based on the content-type of their audio-version-template.  (Probably a loose validation, based on the first param: `video/*`)
- [x] Fix public asset controller to link to non-mp3 files.  Currently it keeps changing the file extension to mp3

cavis:
Working!

PR #348, opened 2017-09-28
Audio callback worker updates

cavis:
See #345.

Also ended up making most of the validations return either a string or nil.  It was a bit problematic to return strings vs `""` vs `[]` vs nil - so simplified the interface a bit.

### DIFF
--- a/app/models/audio_file.rb
+++ b/app/models/audio_file.rb
@@ -51,15 +51,7 @@ class AudioFile < BaseModel
 
   def set_status
     # only do template validations once the audio callback worker has succeeded
-    if status == UPLOADED
-      return
-    elsif status == NOTFOUND
-      self.status_message = "Audio file #{label} not found."
-      return
-    elsif status == FAILED
-      self.status_message = "Audio file #{label} failed to process."
-      return
-    end
+    return if [UPLOADED, NOTFOUND, FAILED].include?(status)
 
     template = audio_version.
                try(:audio_version_template).
--- a/app/models/audio_file.rb
+++ b/app/models/audio_file.rb
@@ -38,6 +38,18 @@ class AudioFile < BaseModel
     [INVALID, VALID, COMPLETE, TRANSFORMING, TRANSFORM_FAILED, TRANSFORMED].include? status
   end
 
+  def enclosure_url
+    is_audio? ? public_url(version: :broadcast, extension: 'mp3') : public_url
+  end
+
+  def enclosure_content_type
+    is_audio? ? 'audio/mpeg' : content_type
+  end
+
+  def is_audio?
+    content_type.present? && content_type.starts_with?('audio/')
+  end
+
   def compliant_with_template?
     status_message.nil?
   end
--- a/app/models/audio_file.rb
+++ b/app/models/audio_file.rb
@@ -53,14 +53,11 @@ class AudioFile < BaseModel
     # only do template validations once the audio callback worker has succeeded
     return if [UPLOADED, NOTFOUND, FAILED].include?(status)
 
-    template = audio_version.
-               try(:audio_version_template).
-               try(:audio_file_templates).
-               try(:find) { |aft| aft.position == position }
+    vtpl = audio_version.try(:audio_version_template)
+    ftpl = vtpl.try(:audio_file_templates).try(:find) { |aft| aft.position == position }
+    errors = vtpl.try(:validate_audio_file, self) || ftpl.try(:validate_audio_file,  self)
 
-    errors = template ? template.validate_audio_file(self) : []
-
-    if errors.empty?
+    if errors.blank?
       self.status_message = nil
       self.status = COMPLETE
     else
--- a/app/models/audio_file.rb
+++ b/app/models/audio_file.rb
@@ -39,14 +39,14 @@ class AudioFile < BaseModel
   end
 
   def enclosure_url
-    is_audio? ? public_url(version: :broadcast, extension: 'mp3') : public_url
+    audio? ? public_url(version: :broadcast, extension: 'mp3') : public_url
   end
 
   def enclosure_content_type
-    is_audio? ? 'audio/mpeg' : content_type
+    audio? ? 'audio/mpeg' : content_type
   end
 
-  def is_audio?
+  def audio?
     content_type.present? && content_type.starts_with?('audio/')
   end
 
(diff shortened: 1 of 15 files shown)


app/models/audio_file.rb
```diff
@@ -51,24 +63,13 @@ def update_version_status
...
-    elsif status == FAILED
-      self.status_message = "Audio file #{label} failed to process."
-      return
-    end
-
-    template = audio_version.
-               try(:audio_version_template).
-               try(:audio_file_templates).
-               try(:find) { |aft| aft.position == position }
+    return if [UPLOADED, NOTFOUND, FAILED].include?(status)
```
cavis: Note: the default `status_message`s are now set in the callback_worker instead of here.  (The worker should be the only place NOTFOUND or FAILED gets set anyways).

app/models/audio_version_template.rb
```diff
@@ -43,42 +43,58 @@ def touch_audio_versions
...
     if length_maximum != 0 && length_minimum > length_maximum
       errors.add(:length_mininum, 'must be less than or equal to length maximum')
     end
   end
+
+  def audio_file_content_type_errors(audio_file)
+    if content_type.present?
+      tpl_type = content_type.split('/').first
+      file_type = audio_file.content_type.try(:split, '/').try(:first)
+      if content_type == 'audio/mpeg' && audio_file.content_type != 'audio/mpeg'
```
cavis: This is a little problematic.  An mp2 has `content_type = 'audio/mpeg'` but `layer = 2`.  But know that the template `content_type` is audio/mpeg isn't actually specific enough to designate layer.  So for now, just letting them pass validation.

app/models/audio_version_template.rb
```diff
@@ -43,42 +43,58 @@ def touch_audio_versions
...
+        "Audio file '#{audio_file.label}' is not an mp3" # very specific on these
+      elsif tpl_type != file_type
+        "Audio file '#{audio_file.label}' is not #{articleize(tpl_type)}"
+      end
+    end
+  end
+
+  def articleize(word)
```
kookster: neat, but is this necessary? could we re-word to  "Audio file '#{audio_file.label}' is not in #{tpl_type} format"

and actually, maybe we should stop saying "Audio file" since videos and other files may come through here - maybe just "File is not in mp3 format"?

app/workers/audio_callback_worker.rb
```diff
@@ -10,39 +10,49 @@ class UnknownAudioTypeError < StandardError; end
...
+                         elsif meta['channels'] == 1
                            SINGLE_CHANNEL
                          end
+
+    # set errors but ignore valid/invalid status (will be determined later)
     if !job['downloaded']
       audio.status = NOTFOUND
-    elsif !job['valid'] || !job['processed']
+      audio.status_message = job['error'] ||
+                             "Audio file for #{audio.label || audio.id} not found"
```
kookster: same note about always calling these audio files.

app/workers/audio_callback_worker.rb
```diff
@@ -10,39 +10,49 @@ class UnknownAudioTypeError < StandardError; end
...
+    elsif !job['processed']
       audio.status = FAILED
+      audio.status_message = job['error'] ||
+                             "Unable to process audio file #{audio.label || audio.id}"
```
kookster: ditto

test/models/audio_version_test.rb
```diff
@@ -98,14 +98,14 @@
...
-      audio_file.update(status: 'not found')
+      audio_file.update(status: 'not found', status_message: 'Was really not found')
       audio_version_with_template.audio_files << audio_file
       audio_version_with_template.update_attributes(explicit: 'explicit')
       audio_version_with_template.status_message.must_include 'not found'
       audio_version_with_template.status.must_equal 'invalid'
       audio_version_with_template.wont_be(:compliant_with_template?)
 
-      audio_file.update(status: 'failed')
+      audio_file.update(status: 'failed', status_message: 'Well that failed to process')
```
kookster: like these messages

test/workers/audio_callback_worker_test.rb
```diff
@@ -16,47 +16,46 @@
...
     defaults = {
       id: audio.id,
+      name: 'test.mp3',
+      size: 1234,
+      mime: 'audio/mpeg',
       downloaded: true,
       valid: true,
       processed: true,
-      name: 'test.mp2'
+      error: nil,
```
kookster: ending with an extra comma? what are they, growing on trees? 🌴

app/models/audio_version_template.rb
kookster: yeah, at some point we'll add more specific validations at a template or distro level, and that will be less of an issue - this is more about getting non-audio/video to work than making audio validations awesome.

(should that check for them matching as opposed to both checked against 'audio/mpeg' string?)

app/models/audio_version_template.rb
```diff
@@ -43,42 +43,58 @@ def touch_audio_versions
...
+    %w(a e i o u).include?(word[0].downcase) ? "an #{word}" : "a #{word}"
```
kookster: regex? `word =~ /^[aeiou]+/i ? "an #{word}" : "a #{word}"`

app/models/audio_version_template.rb
cavis: Not sure i follow on your (2nd comment) there.  You're saying this function should just check that the first half of the mime type matches?  (`audio/` or `video/`)

app/models/audio_version_template.rb
cavis: Hmm... yeah, that "Audio File" thing is annoying, since you see it in the frontend a lot.  Will update.

test/workers/audio_callback_worker_test.rb
cavis: What is this, IE6?

app/models/audio_version_template.rb
kookster: I meant something like:
`if content_type == 'audio/mpeg' && content_type != audio_file.content_type`

since eventually we want to implement for all types, and will become something like
`if content_type != audio_file.content_type`

kookster: very very minor point

cavis: ah, i gotcha.  One minor complication to that ... the current `audio_file.content_type` reflects the content_type of the original file.  Not necessarily any transcoded versions of the file (which may satisfy the audio_version_template).

But transcoding offers many, many complications anyways!

PR #352, opened 2017-10-04
Audio file list position fix

kookster:
- [x] Remove `acts_as_list` from `AudioFile`, leave it on other models
- [x] Add some defaulting of position on `AudioFile` when position is null
- [x] Add validation on `AudioVersion` to make sure audio file positions have no gaps

### DIFF
--- a/app/models/audio_version.rb
+++ b/app/models/audio_version.rb
@@ -64,6 +64,13 @@ class AudioVersion < BaseModel
       return
     end
 
+    pos = 0
+    if audio_files.detect { |af| pos += 1; af.position != pos }
+      self.status = INVALID
+      self.status_message = "Audio file missing for position #{pos}"
+      return
+    end
+
     errors = audio_version_template.try(:validate_audio_version, self)
     if errors.blank?
       self.status = COMPLETE
--- a/app/models/audio_version.rb
+++ b/app/models/audio_version.rb
@@ -65,7 +65,7 @@ class AudioVersion < BaseModel
     end
 
     pos = 0
-    if audio_files.detect { |af| pos += 1; af.position != pos }
+    if audio_files.detect { |af| af.position != (pos += 1) }
       self.status = INVALID
       self.status_message = "Audio file missing for position #{pos}"
       return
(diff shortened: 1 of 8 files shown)


app/models/audio_version.rb
```diff
@@ -64,6 +64,13 @@ def set_status
       return
     end
 
+    pos = 0
+    if audio_files.detect { |af| af.position != (pos += 1) }
+      self.status = INVALID
+      self.status_message = "Audio file missing for position #{pos}"
```
cavis: It doesn't actually matter.  But I might prefer this error be _after_ the audio-version-template errors in priority.  Meaning you'd get the current "Audio version has 2 files but must have 3 files" error before this one.

test/factories/audio_version_factory.rb
```diff
@@ -15,7 +15,8 @@
     after(:create) do |audio_version, evaluator|
       c = evaluator.audio_files_count
       c = (rand(5) + 1) if (evaluator.audio_files_count == 0)
-      FactoryGirl.create_list(:audio_file, c, audio_version: audio_version)
+      c = 0 if evaluator.audio_files_count < 0
```
cavis: So now < 0 means 0, and 0 means > 0?

app/models/audio_version.rb
kookster: thanks, I think you're right - I'll move it.

test/factories/audio_version_factory.rb
kookster: just checked, no place in the code is using the current 0 value, so getting rid of the random, and simplifying this.

app/models/audio_version.rb
```diff
@@ -72,5 +72,12 @@ def set_status
       self.status = INVALID
       self.status_message = errors
```
cavis: You probably need a `return` here now, right?

kookster: argh, yes!

kookster:
@cavis that should do it

cavis:
:+greatest:

PR #358, opened 2017-10-18
Strip tags on import

gcampo88:
When a feed had categories like so: 
```RSS
<category>
  Something
</category>
<category>
  Something else
</category>
```
the tags were getting saved with newlines. Which looks pretty weird on Publish!

### DIFF
--- a/app/models/podcast_import.rb
+++ b/app/models/podcast_import.rb
@@ -250,7 +250,7 @@ class PodcastImport < BaseModel
       title: clean_string(entry[:title]),
       short_description: clean_string(short_desc(entry)),
       description_html: entry_description(entry),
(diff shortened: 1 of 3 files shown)


app/models/podcast_import.rb
```diff
@@ -250,7 +250,7 @@ def create_story(entry, series)
       title: clean_string(entry[:title]),
       short_description: clean_string(short_desc(entry)),
       description_html: entry_description(entry),
-      tags: entry[:categories],
+      tags: Array(entry[:categories]).map(&:strip),
```
kookster: should we also remove nil/empty elements?
`Array(entry[:categories]).map(&:strip).reject(&:blank?)`

gcampo88:
@kookster ok to merge?

Issue #136, opened 2016-04-20
Zoom param not working


cavis:
At some point, it seems like a Roar upgrade broke the `?zoom` param.

kookster:
probably not terrible to track down, but no app is using zoom, so not a high priority

kookster:
https://github.com/PRX/hal_api-rails/pull/9

PR #360, opened 2017-10-26
Fix zoom

kookster:
- [x] Uses a new version of `hal_api-rails` https://github.com/PRX/hal_api-rails/pull/9
- [x] Update paged collection representer to use a concern
- [x] Use `HalApi::RepresentedModel`, not the one in the app
- [x] Use `included` methods on controller to efficiently load some resources


kookster:
#136

app/controllers/api/base_controller.rb
```diff
@@ -67,6 +67,17 @@ def authorization
     Authorization.new(prx_auth_token) if prx_auth_token
   end
 
+  def filtered(arel)
+    keys = self.class.resources_params || []
+    where_hash = params.slice(*keys)
+    if where_hash.key?('story_id')
+      where_hash['piece_id'] = where_hash.delete('story_id')
```
cavis: I like this better than the fix that was in the gem!

app/representers/api/paged_collection_representer.rb
```diff
@@ -1,63 +1,5 @@
-# encoding: utf-8
+require 'hal_api/representer/collection_paging'
 
 class Api::PagedCollectionRepresenter < Api::BaseRepresenter
```
cavis: Hmm... doesn't the HalApi already include the equivalent of this representer?

kookster: similar, but wrong app specific curies, so need to define our own - the `Api::BaseRepresenter` provides that

cavis: makes sense.

PR #361, opened 2017-10-29
Improve API performance

kookster:
- [x] Update `hal_api-rails` to use query cache for unpaged embedded resources
- [x] Ad `includes` for a story's embedded resources
- [x] zoom less by default
- [x] zoom audio for stories w/o paging

### DIFF
--- a/app/controllers/api/stories_controller.rb
+++ b/app/controllers/api/stories_controller.rb
@@ -89,13 +89,13 @@ class Api::StoriesController < Api::BaseController
       { promos: [:audio_files] },
       { account: [:image, :address, { opener: [:image] }] },
       { series: [:images, :account, :audio_version_templates] },
-      { distributions: [:audio_version_templates]},
       :creator,
       :images,
       :license,
       :topics,
       :tones,
       :formats,
+      :distributions,
       :user_tags
     )
   end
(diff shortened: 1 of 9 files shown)


app/controllers/api/stories_controller.rb
```diff
@@ -86,13 +86,16 @@ def filtered(resources)
   def included(relation)
     relation.includes(
       { audio_versions: [:audio_files] },
```
gcampo88: could also include :audio_version_templates for distribution, and :distributions for series?

kookster: hmm, I could add the distro av templates, those are included by default in the embedded distro, but the min version of series embedded in a story does not include distributions.

gcampo88: it links to distributions, no? (not embedding). does the `count` matter?
```ruby 
# app/representers/api/min/series_representer.rb

  link :distributions do
    {
      href: api_series_distributions_path(represented),
      count: represented.distributions.count
    } if represented.id
  end
```

kookster: also, this is the story distribution, not the series distro, so there are no templates associated with it.

PR #363, opened 2017-11-01
WIP Better ordered announcements

kookster:


app/controllers/concerns/announce_actions/announce_filter.rb
```diff
@@ -19,7 +19,13 @@ def after(controller)
       model = announce_resource(action, controller)
       subject = announce_subject(model)
       message = announce_message(model)
-      announce(subject, announce_action, message)
+      if model.respond_to?(:with_lock)
```
kookster: In tests at least, this is not always an AR model, so handle when this isn't actually lockable.

app/workers/audio_callback_worker.rb
```diff
@@ -9,6 +9,17 @@ class UnknownAudioTypeError < StandardError; end
 
   def perform(_sqs_msg, job)
     audio = AudioFile.find(job['id'])
+    update_audio(audio, job)
```
kookster: I was also playing with putting this update method inside the lock block as well - I may still do that, but in any case, I find this more readable to separate the methods.

cavis: Yeah, it's a bit of an edge case to get 2 audio callbacks at once.  Then again, this whole debacle is about edge cases.

app/workers/audio_callback_worker.rb
```diff
@@ -9,6 +9,17 @@ class UnknownAudioTypeError < StandardError; end
...
+    story = audio.story
+    story.with_lock do
```
cavis: And you confirmed that `with_lock` reloads the model?

gcampo88: `lock!` reloads the record, per [docs](http://api.rubyonrails.org/classes/ActiveRecord/Locking/Pessimistic.html), and `with_lock` uses `lock!` (from lurker gigi)

kookster: yes, I looked up the code, it essentially does `model.lock!; model.reload; yield model` inside a transaction.

kookster:
the audio and image updates are not synched to feeder, they are used for lambda processing, so updating those announcements doesn't affect the feeder synch issue

cavis:
+merge!

Issue #353, opened 2017-10-04
handle new itunes tags on import 


gcampo88:
If the feed we're importing includes the [new iOS11 tags](http://podcasts.apple.com/resources/spec/ApplePodcastsSpecUpdatesiOS11.pdf) then we should maintain those when we import it.

PR #364, opened 2017-11-06
Add new apple podcasts tags to import

gcampo88:
#353

### DIFF
--- a/app/models/podcast_import.rb
+++ b/app/models/podcast_import.rb
@@ -144,6 +144,7 @@ class PodcastImport < BaseModel
     podcast_attributes[:complete] = (clean_string(feed.itunes_complete) == 'yes')
     podcast_attributes[:copyright] ||= clean_string(feed.media_copyright)
     podcast_attributes[:keywords] = parse_keywords(feed)
+    podcast_attributes[:serial_order] = feed.itunes_type.downcase == 'serial'
 
     self.podcast = distribution.add_podcast_to_feeder(podcast_attributes)
     podcast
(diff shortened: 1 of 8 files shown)


app/models/podcast_import.rb
```diff
@@ -339,6 +343,7 @@ def create_episode(story, entry)
     create_attributes[:keywords] = (entry[:itunes_keywords] || '').split(',').map(&:strip)
     create_attributes[:position] = entry[:itunes_order]
     create_attributes[:url] = episode_url(entry) || distro.default_url(story)
+    create_attributes[:itunes_type] = entry[:itunes_episode_type] unless entry[:itunes_episode_type].blank?
```
cavis: Why the blank check here, but not in the other fields?

test/models/podcast_import_test.rb
```diff
@@ -226,4 +231,8 @@ def stub_requests
     with(body: /prxUri/,
          headers: { 'Authorization' => 'Bearer thisisnotatoken' }).
     to_return(status: 200, body: json_file('transistor_episode'), headers: {})
+
+stub_request(:get, "https://feeder.prx.org/api/v1/authorization/episodes/153e6ea8-6485-4d53-9c22-bd996d0b3b03").
```
cavis: Why is this added now?

app/models/podcast_import.rb
gcampo88: no particular reason, really -- just that the other fields were inside a call to `create`

test/models/podcast_import_test.rb
gcampo88: none of the existing tests check the Feeder episode created for a story, so I had to stub that method to add this one on line 97: `f.distributions.first.get_episode.itunes_type.must_equal 'full'`

cavis:
Makes good sense.  :+7: :participation_trophy_medal:

gcampo88:
participation!

PR #378, opened 2017-12-22
Don't set path on podcast during import

gcampo88:


pkarman:
@gcampo88 feel free to merge.

PR #384, opened 2018-01-12
Handle when public asset has no file name extension

kookster:
Found this problem importing `/stream` urls from soundcloud.
Need to handle when there is no extension, and so make that an optional part of the route matcher.

However, for incoming requests, parse out the extension as there is not a good regex way to have the router differentiate between the extension and the base filename.

### DIFF
--- a/app/models/concerns/public_asset.rb
+++ b/app/models/concerns/public_asset.rb
@@ -24,6 +24,11 @@ module PublicAsset
     n = o[:name]
     x = o[:extension]
 
+    if x.blank? && n.include?('.')
+      n = get_basename(o[:name])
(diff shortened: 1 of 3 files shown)


app/models/concerns/public_asset.rb
```diff
@@ -24,6 +24,11 @@ def public_url_token(options={})
     n = o[:name]
     x = o[:extension]
 
+    if x.blank? && n.include?('.')
```
gcampo88: You were starting to explain this in person before, but I can't remember exactly the reasoning -- wherefore `include?('.')`? are we looking for cases other than n = `filename.` ?

Issue #387, opened 2018-01-18
What happens to existing stories when I make a Series a Podcast? 


gcampo88:
We just had an issue where a Publish user created a Series, created a Story within that Series, then created a Podcast for the series, and expected the story to yield an episode within that podcast. That didn't happen, because the episode distribution for a story only gets called on the story create action. We should figure out how best to handle this case. Ideas: on story update, check whether any of its `audio_versions` belong to an `audio_version_template` that goes with a podcast distribution, and if so, create the ep distribution to go along with it.

PR #388, opened 2018-01-18
Create episode distribution on story update if it seems necessary

gcampo88:
https://github.com/PRX/cms.prx.org/issues/387

### DIFF
--- a/app/controllers/api/stories_controller.rb
+++ b/app/controllers/api/stories_controller.rb
@@ -18,19 +18,22 @@ class Api::StoriesController < Api::BaseController
   end
 
   def after_update_resource(res)
-    dist_tmpl = res
-      .try(:series)
-      .try(:distributions)
-      .try(:find) { |d| d.kind == "podcast" }
-      .try(:audio_version_template_id)
+    series_dists = res.try(:series).try(:distributions)
 
-    version_tmpl = res.try(:audio_versions)
-      .try(:find) { |av| !av.audio_version_template_id.nil? }
-      .try(:audio_version_template_id)
+    story_template_ids = res
+      .try(:audio_versions)
+      .try(:map) { |av| av.audio_version_template_id }
+      .try(:compact)
 
-    if dist_tmpl && version_tmpl && dist_tmpl == version_tmpl && res.distributions.empty?
-      res.create_story_distributions
+    if series_dists && story_template_ids
+      missing_dists = series_dists.select do |series_dist|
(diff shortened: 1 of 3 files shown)


app/controllers/api/stories_controller.rb
```diff
@@ -17,6 +17,22 @@ def after_create_resource(res)
     res.create_story_distributions
   end
 
+  def after_update_resource(res)
+    dist_tmpl = res
+      .try(:series)
+      .try(:distributions)
+      .try(:find) { |d| d.kind == "podcast" }
```
kookster: could there be more than one? I mean, I think the models support it, even if we don't do that on the UI.
I think this works, but I wonder if it would be good to get a list in dist_tmpl of ids

kookster: as I think about it, I think this logic of catching up on missing distributions is not specific to podcasts - we just don't have other distros at the moment, so I wonder if filtering on "podcasts" is necessary - maybe we should do this for all series distributions?

app/controllers/api/stories_controller.rb
```diff
@@ -17,6 +17,22 @@ def after_create_resource(res)
...
+      .try(:audio_version_template_id)
+
+    version_tmpl = res.try(:audio_versions)
+      .try(:find) { |av| !av.audio_version_template_id.nil? }
```
kookster: this may be bad feedback (and not a demand for change) as what you've done here is quite readable, but if you did have a list of `dist_tmpl` ids, then you could check here if `dist_tmpl.include?(av.audio_version_template_id)` and save yourself that comparison below.

app/controllers/api/stories_controller.rb
```diff
@@ -17,6 +17,22 @@ def after_create_resource(res)
...
+      .try(:audio_version_template_id)
+
+    if dist_tmpl && version_tmpl && dist_tmpl == version_tmpl && res.distributions.empty?
+      res.create_story_distributions
```
kookster: this method is going to make distros for all distributions that are and are not podcasts - you've prevented it making a dupe podcast episode distro above, but would still create other dupe distros for any other kind of series distro

gcampo88:
good feedback @kookster ! Second attempt is ready for you now.

kookster:
looks really good! Like how you refactored the story model a bit too.

Do we need a test for when there is an existing story distribution, to make sure it doesn't try to create a dupe on update? Or does one of the tests cover that, and I missed it?

PR #390, opened 2018-01-20
Apply defaults from model when generating urls, not validating

kookster:


kookster:
@gcampo88 no rush on this (pls wait till Mon to review), just meant to get to it on Friday, and so I finished it while I could now.

Issue #391, opened 2018-01-23
Import fails on emoji in feed


kookster:
mysql 5.0 can't handle 4 byte utf8 column type, so, need to keep these out for now.

Ideas:
https://github.com/maximeg/activecleaner/blob/master/lib/active_cleaner/utf8mb3_cleaner.rb#L38
https://stackoverflow.com/questions/16487697/how-to-remove-4-byte-utf-8-characters-in-ruby
https://gist.github.com/adamlwatson/9623703

kookster:
(estimate of 2 b/c of time to research solutions)

pkarman:
When I faced this issue before I converted the 4-byte characters into ASCII entity expressions: &#nnnnn; e.g. That's only if you care about preserving them.

PR #392, opened 2018-01-23
Filter out 4 byte utf8 characters

kookster:
#391


app/models/concerns/import_utils.rb
```diff
@@ -57,6 +58,8 @@ def clean_text(text)
     return nil if text.blank?
     result = remove_feedburner_tracker(text)
     sanitize_html(result)
```
gcampo88: you need a `result = sanitize_html(result)` here for this to work

kookster:
sanitize 🗑 is back!

PR #393, opened 2018-01-23
Publish watcher for story distributions

kookster:
A new scheduled job that runs every few minutes to check if recently published episodes in cms have been distributed and published on feeder* (the `belt` is messages announced on changes, this is the `suspenders` to make sure those messages work).

Also will send errors to newrelic when these are not distributed when they should be, or not published when they should be, as well as sending errors to NR if they fail to be updated.

Relies on DB change to have cms say_when tables to store the scheduled recurring job:
https://github.com/PRX/exchange.prx.org/pull/339

(*) the code actually doesn't care how the story is distributed, written to test if the distribution, of whatever kind, is complete


app/jobs/application_job.rb
```diff
@@ -1,4 +1,15 @@
-# encoding: utf-8
+require 'newrelic_rpm'
 
 class ApplicationJob < ActiveJob::Base
+  around_perform :use_db_pool
+
+  rescue_from(StandardError) do |e|
```
pkarman: If I'm reading this correctly, this base class now surfaces all exceptions to New Relic? Nice!

app/models/distribution.rb
```diff
@@ -9,6 +9,45 @@ class Distribution < BaseModel
   has_many :audio_version_templates, through: :distribution_templates
   serialize :properties, HashSerializer
 
+  def self.check_published!(data = nil)
+    data ||= {}
+    from_secs = (data[:from_secs] || '60').to_i
```
pkarman: a nit: `from` and `to` are a little confusing for me here as variables names. You're defining a range window within which you want to pull recently published stories? Maybe `start` and `end` would be clearer? I find naming hard when doing Time math, so maybe it's just me who is confused.

app/models/distribution.rb
```diff
@@ -9,6 +9,45 @@ class Distribution < BaseModel
...
+        end
+
+        if !dist.published?
+          notice_message("Story #{story_id} distribution not published: #{dist.url}")
+          dist.publish!
+        end
+      end
+    end
+  rescue StandardError => e
+    NewRelic::Agent.notice_error(e, custom_params: { story: story_id, distribution: dist_id })
```
pkarman: would it be helpful to have the time range in the NR log too?

app/models/distribution.rb
```diff
@@ -9,6 +9,45 @@ class Distribution < BaseModel
...
+  end
+
+  def self.notice_message(msg)
+    NewRelic::Agent.notice_error(RuntimeError.new(msg))
+  end
+
+  def self.recently_published_stories(from_secs, to_secs)
+    Story.where(
+      '`published_at` <= ? AND `published_at` >= ?',
+      from_secs.seconds.ago,
```
pkarman: so here I can see how `from` and `to` are used. I am translating mentally to `from` == `end` and `to` == `start`.

app/models/distributions/podcast_distribution.rb
```diff
@@ -1,16 +1,44 @@
...
+
+    if published_at_s = podcast.attributes['published_at']
+      published_at = DateTime.parse(published_at_s)
+      published = published_at <= DateTime.now
+    end
+    published && stories_published?
+  end
+
+  def stories_published?
+    story_distributions.all? { |sd| sd.distributed? && (!sd.story.published? || sd.published?) }
```
pkarman: this boolean logic is a bit tricky - the story is *not* published or the distribution *is* published?

app/jobs/application_job.rb
kookster: that's the idea

app/models/distribution.rb
kookster: the range shouldn't really change, so I didn't think it would help much

app/models/distribution.rb
kookster: `end` is a reserved word in ruby so I try to avoid that, but I could use `start_time` and `end_time`?

app/jobs/application_job.rb
```diff
@@ -1,4 +1,15 @@
-# encoding: utf-8
+require 'newrelic_rpm'
 
 class ApplicationJob < ActiveJob::Base
+  around_perform :use_db_pool
```
gcampo88: nice!

app/models/distributions/podcast_distribution.rb
kookster: if the story isn't published, then no need to check the distro.
if it is published, then the distro needs to be published to.

app/jobs/application_job.rb
kookster: gonna update feeder too!

app/models/distribution.rb
kookster: ok, changed

app/models/distributions/podcast_distribution.rb
kookster: added that as a comment in the code

pkarman:
variable rename looks good - thanks

pkarman:
@gcampo88 that's what I mean about Time math naming being hard - are we oriented toward moving forward or backward in time? In this case it's a window slice. I can be confused either way.

gcampo88:
cool. I'm good with these names @kookster + @pkarman

kookster:
ok - I have deploys fixed - and production exchange deployed and db migrated, so this is ready to go!

PR #395, opened 2018-01-24
Filter out 4 byte utf 8 before saving import entry

kookster:
Prior changes correctly filter out values saved to story models, but did not filter out the 4byte values before saving serialized hash of feed entry to the import

### DIFF
--- a/app/models/concerns/import_utils.rb
+++ b/app/models/concerns/import_utils.rb
@@ -62,7 +62,11 @@ module ImportUtils
     return nil if text.blank?
     result = remove_feedburner_tracker(text)
     result = sanitize_html(result)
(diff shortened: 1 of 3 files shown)


app/models/concerns/import_utils.rb
```diff
@@ -62,7 +62,11 @@ def clean_text(text)
     return nil if text.blank?
     result = remove_feedburner_tracker(text)
     result = sanitize_html(result)
-    result.each_char.select { |char| char.bytesize < 4 }.join('')
+    remove_utf8_4byte(result)
```
gcampo88: this is where elixir would be nice! `|>` ftw

PR #397, opened 2018-02-13
Import episodes only

kookster:
- [x] Add `config[:episodes_only]` to indicate the series and podcast are already created, and should not be updated
- [x] Update retrieve and validate the series and podcast when `:episodes_only` is true

### DIFF
--- a/test/models/podcast_import_test.rb
+++ b/test/models/podcast_import_test.rb
@@ -6,6 +6,7 @@ describe PodcastImport do
 
   let(:user) { create(:user) }
   let(:account) { create(:account, id: 8, opener: user) }
+  let(:series) { create(:series, account: account) }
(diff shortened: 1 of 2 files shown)


test/models/podcast_import_test.rb
```diff
@@ -90,6 +91,31 @@
...
+
+    it 'must have podcast distribution' do
+      series.distributions.delete_all
+      importer.series = series
+      exception = -> { importer.import }.must_raise(RuntimeError)
+      exception.message.must_be :start_with?, 'No podcast distribution'
+    end
+
+    it 'imports with a series and podcast' do
+      # series.distributions << create(:podcast_distribution, distributable: series)
```
gcampo88: do you need this commented out line?

kookster: nope, needed it before I broke out this test, will `rm`

PR #406, opened 2018-02-21
Allow container db services to co-exist with host services

pkarman:
I was already running pg and mysql on the host, so the container instances complained about the port already being in use.

Not sure if/how this might break expectations folks have about workflows. All tests pass locally for me.

NOTE this depends on #405 so I will rebase once that is merged.


kookster:
yeah - @pkarman looks like you need to merge master

cavis:
👍

pkarman:
thanks @cavis @kookster - branch rebased to a single commit now

Issue #344, opened 2017-09-20
get rid of unused importer code


gcampo88:
lib/podcast_importer is not going to be used; get rid of it.

PR #409, opened 2018-02-26
Set content_type on audio_version_template import

pkarman:


pkarman:
There did not seem to be any existing tests for specific attribute values on the template. Happy to add some if that seems worthwhile.

app/models/audio_version_template.rb
```diff
@@ -87,7 +87,8 @@ def audio_file_content_type_errors(audio_file)
     if content_type.present?
       tpl_type = content_type.split('/').first
       file_type = audio_file.content_type.try(:split, '/').try(:first)
-      if content_type == 'audio/mpeg' && audio_file.content_type != 'audio/mpeg'
+      if content_type == AudioFile::MP3_CONTENT_TYPE && \
```
kookster: I don't think you need that backslash, the `&&` is enough to indicate to the interpreter that the expression continues to the next line.

kookster:
Yeah, @gcampo88 is right, we need to handle when there is a non-mp3 file in the import (usually video).

It's fine for `feeder_importer.rb` to go ahead and create the default audio version template with mp3 as the content type, but we also need a method to check for a given episode, if the audio for that episode is mp3, and if it is not, create a new template  for the series with the correct content type, so that can be used.

This also needs to be done in the regular podcast import (not just from feeder), where there is logic to create a new version if the number of files is different - but it doesn't take an arg or consider if the content type is wrong:
https://github.com/PRX/cms.prx.org/blob/master/app/models/podcast_import.rb#L300

pkarman:
thanks @gcampo88 
I took care of #344 in #411 
I added logic to add the video template if any enclosure references video content. I suppose this assumes that we always have audio and sometimes, additionally, video, but that seems like a safe assumption.
what do you think?

test/lib/feeder_importer_test.rb
```diff
@@ -67,10 +67,11 @@
     story.app_version.must_equal PRX::APP_VERSION
     story.creator_id.must_equal user_id
     story.account_id.must_equal account_id
-    story.title.must_equal 'No Inoculation without Representation!'
-    story.short_description.must_equal 'A tale of vaccinations and the American Revolution'
-    story.description_html.must_match /Vaccinations, in one form or another/
-    story.tags.must_equal ['Adams', 'American Revolution', 'inoculation', 'transistor', 'vaccine']
+
```
pkarman: I changed these to reference the original episode rather than hard-coded strings, since our fixtures order may change over time (as they have now with #410).

app/models/podcast_import.rb
```diff
@@ -332,6 +336,12 @@ def get_or_create_template(segments)
     template
   end
 
+  def podcast_label(contains_video, num_segments)
```
gcampo88: nice

gcampo88:
Lookin better! Last things I see for now are for the feeder import script, to check within the `create_story` method whether the episode `contains_video?`, and if so, to use the video audio_version_template as the template when creating audio_versions for that story. (currently it's using the attr_accessor `template` which will always be the audio template)
``` 
# feeder_importer.rb line 215   

 version = story.audio_versions.create!(
      audio_version_template: template, # <- here
      label: 'Podcast Audio',
      explicit: episode.explicit
  )
```
Ah -- and you'll want to make sure the PodcastDistribution knows about the video audio_version_template as well (a distribution `has_many` audio_version_templates, in addition to having an `audio_version_template_id`, from before we started allowing shows to have multiple structures) -- see line 178 in feeder_importer.rb. Right now it only knows about the audio template. 

Sorry, this ticket is a bit bigger than originally thought!

pkarman:
@gcampo88 ok I think I've addressed the issues you identified.

pkarman:
@kookster would appreciate your 👀 on this too since it touches so many things.

lib/feeder_importer.rb
```diff
@@ -139,29 +143,50 @@ def create_series
...
+          length_maximum: 0
+        )
+      end
+    end
+
     self.distribution = Distributions::PodcastDistribution.create!(
       url: "#{feeder_root}/podcasts/#{podcast_id}",
       distributable: series,
-      audio_version_template: template
+      audio_version_template: audio_template
```
kookster: you can set both templates in there now - it used to support only one, but a distro can now have multiple templates.

lib/feeder_importer.rb
```diff
@@ -139,29 +143,50 @@ def create_series
...
     )
+    if contains_video
+      DistributionTemplate.create(distribution: distribution, audio_version_template: video_template)
```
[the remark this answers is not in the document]
kookster: yeah, you can do this on the distro create, not sure you need to do by itself, or you could add each templates this way.

pkarman:
@kookster is something like 46af4ce what you had in mind?

PR #411, opened 2018-02-27
Drop lib/podcast_importer 

pkarman:
Not used - see instead `app/model/podcast_importer`


kookster:
sigh - yeah - it's an interesting approach, but if I ever want it back, I can find it the git history.

Issue #399, opened 2018-02-15
Set podcast `locked` attribute on import


gcampo88:
per https://github.com/PRX/feeder.prx.org/pull/278 
podcasts can now be `locked`, which means they won't publish new RSS to s3. Should set this attribute to true at the beginning of an import. Can then either set it to false at the end of import as part of import process, or leave that as a manual step that the importer (human) can do after QA.

PR #413, opened 2018-03-01
Lock podcast during import

gcampo88:
#399


gcampo88:
good call @cavis. Back to you!

cavis:
:+all:

PR #418, opened 2018-03-12
Apply same description logic as episode_import model

pkarman:
The `EpisodeImport` model uses an array of "description" fields to populate the Episode.description on import. Apply the same logic to the `FeederImporter`.

### DIFF
--- a/lib/feeder_importer.rb
+++ b/lib/feeder_importer.rb
@@ -213,7 +213,7 @@ class FeederImporter
   end
 
   def episode_description(episode)
-    attrname = %i(:content :summary :description :subtitle :title).find { |d| !episode[d].blank? }
+    attrname = %i(content summary description subtitle title).find { |d| !episode[d].blank? }
     episode[attrname]
   end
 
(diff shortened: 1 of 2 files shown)


test/lib/feeder_importer_test.rb
```diff
@@ -61,6 +61,9 @@
     now = Time.now
     episode.published_at = now
 
+    # tweek description so we verify we get content
```
gcampo88: tweak!

lib/feeder_importer.rb
```diff
@@ -212,6 +212,11 @@ def template_for_episode(episode)
     episode.contains_video? ? video_template : audio_template
   end
 
+  def episode_description(episode)
+    attr = [:content, :summary, :description, :subtitle, :title].find { |d| !episode[d].blank? }
```
gcampo88: is `attr` a reserved word in ruby and/or rails? [some](http://www.rubymagic.org/posts/ruby-and-rails-reserved-words) [sleuthing](http://reservedwords.herokuapp.com/) points to no, but we avoided using it in [the similar method](https://github.com/PRX/cms.prx.org/blob/master/app/models/episode_import.rb#L174) in EpisodeImport.rb, and Atom and Github are both giving it special syntax highlighting!

pkarman:
Thanks @gcampo88 - think I hit your comments.

PR #429, opened 2018-06-26
Elasticsearch

pkarman:
Basic infrastructure for building ES indices via rake task. One API endpoint partially implemented for Story model.

Still TODO via Epic #444
* [ ] Re-implement API filters against ES index for Story #441
* [ ] Re-implement Networks search for non-Story models #442
* [ ] Incremental updates #443

### DIFF
--- a/app/controllers/api/stories_controller.rb
+++ b/app/controllers/api/stories_controller.rb
@@ -94,7 +94,9 @@ class Api::StoriesController < Api::BaseController
 
   def filtered(resources)
     resources = resources.v4 if filters.v4?
-    resources = resources.match_text(filters.text) if filters.text?
+    Rails.logger.debug("before: #{resources}")
+    resources = resources.match_text(filters.text, params, current_user) if filters.text?
+    Rails.logger.debug("after: #{resources}")
     if highlighted?
       resources
     else
--- a/app/controllers/api/stories_controller.rb
+++ b/app/controllers/api/stories_controller.rb
@@ -13,6 +13,11 @@ class Api::StoriesController < Api::BaseController
 
   announce_actions :create, :update, :destroy, :publish, :unpublish
 
+  def search
+    @stories = Story.match_text(params[:q], params, current_user)
+    index
+  end
+
   def after_create_resource(res)
     res.create_story_distributions
   end
--- a/app/controllers/api/stories_controller.rb
+++ b/app/controllers/api/stories_controller.rb
@@ -99,9 +99,7 @@ class Api::StoriesController < Api::BaseController
 
   def filtered(resources)
     resources = resources.v4 if filters.v4?
-    Rails.logger.debug("before: #{resources}")
     resources = resources.match_text(filters.text, params, current_user) if filters.text?
-    Rails.logger.debug("after: #{resources}")
     if highlighted?
       resources
     else
--- a/app/controllers/api/stories_controller.rb
+++ b/app/controllers/api/stories_controller.rb
@@ -14,7 +14,7 @@ class Api::StoriesController < Api::BaseController
   announce_actions :create, :update, :destroy, :publish, :unpublish
 
   def search
-    @stories = Story.match_text(params[:q], params, current_user)
+    @stories = Story.text_search(params[:q], params, current_user)
     index
   end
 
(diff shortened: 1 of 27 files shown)


app/models/concerns/searchable.rb
```diff
@@ -0,0 +1,72 @@
...
+    def remove_from_index
+      __elasticsearch__.destroy_document
+      refresh_index
+    end
+
+    def refresh_index
+      self.class.__elasticsearch__.refresh_index!
+    end
+
+    def self.rebuild_index
```
cavis: What is this used for?  (Not called anywhere in this PR).  Is this the high-availability async reindexing?

app/services/story_query_builder.rb
```diff
@@ -0,0 +1,192 @@
...
+    if clauses.length == 2
+      clauses.map { |c| "(#{c})" }.join(" AND ")
+    elsif clauses.length == 1
+      clauses[0].to_s
+    else
+      ""
+    end
+  end
+
+  def munge_fielded_params(fielded)
```
cavis: What is this used for?  Some sort of created-in-range query that we don't use yet?

app/services/story_query_builder.rb
```diff
@@ -0,0 +1,192 @@
...
+  end
+
+  def published_filter
+    searchdsl = self
+    Filter.new do
+      term published: true
+    end
+  end
+
+  def authz_filter
```
cavis: Guessing that a lot of these methods will eventually be shared with other types of query too.  (all the other models where you `include Searchable`)

bin/application
```diff
@@ -29,6 +29,20 @@ WaitForMysql () {
   done
 }
 
+WaitForElasticsearch () {
```
cavis: I see how this makes sense for the tests.  But I wonder if ES-availability is really necessary for the CMS app to run at all.  (I'm always hesitant to add app-startup dependencies)

cavis: Even in dev, I can see it being handy to be able to start CMS without ES.  The search endpoints just won't work.

config/routes.rb
```diff
@@ -16,6 +16,7 @@
 
       resources :stories, except: [:new, :edit] do
         get 'random', on: :collection
+        get 'search', on: :collection
```
cavis: Hrmm... yeah, remind me what this is for, if we're just overriding the `?text` filter.

docker-compose.yml
```diff
@@ -57,3 +59,18 @@ pg:
     - "5432"
   volumes:
     - ./test/db/:/docker-entrypoint-initdb.d/
+elasticsearch:
+  image: docker.elastic.co/elasticsearch/elasticsearch-oss:6.3.0
+  container_name: elasticsearch
+  environment:
+    - cluster.name=docker-cluster
+    - bootstrap.memory_lock=true
+    - "ES_JAVA_OPTS=-Xms512m -Xmx512m"
```
cavis: Hmm... maybe these other ENVs are no longer necessary, if this changes to single-node mode.  (And should change this one to match the docker-compose-ci)

app/controllers/api/stories_controller.rb
```diff
@@ -94,7 +99,9 @@ def resources_base
 
   def filtered(resources)
     resources = resources.v4 if filters.v4?
-    resources = resources.match_text(filters.text) if filters.text?
+    Rails.logger.debug("before: #{resources}")
```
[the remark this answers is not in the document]
pkarman: no. I'll zap.

app/services/story_query_builder.rb
pkarman: yes. I basically lifted a query builder class I built a couple years ago and started adapting. There are several vestigial pieces. I will mark those so it's clear what isn't (yet) used.

docker-compose.yml
pkarman: yeah I think I cargo-culted most of the env stuff from an ES example.

pkarman:
@cavis I commented and reverted as much as I could find, and added tests for the new ES functionality.

Since this PR just implements the basic infrastructure and I didn't want it to get too big, I created a number of other tickets to describe what I think are the TODOs. See #444 

I will try adding some unit tests for the DSL query builder. We can wait on those for this PR, or merge in a subsequent PR.

bin/application
```diff
@@ -70,6 +85,7 @@ ApplicationRun () {
 
   UpdateEnvironment
   WaitForMysql
+  WaitForElasticsearch
```
cavis: (I think maybe you meant to take this out)

config/routes.rb
cavis: +1 on separate endpoint, at least to start.

PR #435, opened 2018-08-02
Feature/upgrade docker compose yml

svevang:
This PR bumps the major version of the docker-compose.yml file to "2" which allows the addition of a `networks` top-level key.  Based on this updated version I was able to get a dev environment running in a WIP over here, where cms, id, and exchange all talk to each other: https://github.com/PRX/cms.prx.org/compare/feature/cms-id-exchange-shared-database

Also, this changes the port exposed on the docker host from 3000 to 3001.  This allows multiple docker-compose rails apps to boot up at the same time. If this PR is merged, I'll open similar PRs over in id, exchange etc, bumping those ports as well (3002, 3003, etc.). The effect of this should be pretty minimal, since the dinghy proxy hides these ports under the virtual host (e.g. cms.prx.docker)

### DIFF
--- a/config/database.yml
+++ b/config/database.yml
@@ -30,7 +30,6 @@ makara: &makara
 
 development:
   <<: *default
-  database: cms_development
 
 test:
   <<: *default
--- a/docker-compose-ci.yml
+++ b/docker-compose-ci.yml
@@ -56,3 +56,5 @@ services:
       - esdata1:/usr/share/elasticsearch/data
     ports:
       - 9200:9200
+volumes:
+  esdata1:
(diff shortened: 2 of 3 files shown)


svevang:
@pkarman Thanks for taking a look.  Hopefully this would simplify your overlay compose file. I imagine running these services simultaneously is fairly typical, so made sense to resolve the port collisions. Especially, if we are using the dinghy request proxy (or similar) to resolve proper names to the services, than the oddly numbered dev ports here are papered over.

If we decide to go ahead with this change, then there would be a similar Pull Request in the feeder/id/exchange repos.

PR #438, opened 2018-08-10
Squash a couple errors

cavis:


app/models/concerns/fixerable.rb
```diff
@@ -137,13 +137,14 @@ def fixerable_final_path
 
   # temporary location
   def fixerable_temp_url(options = {})
-    uri = URI.parse(fixtemp)
+    value = fixtemp.include?('://') ? fixtemp : "s3://#{ENV['AWS_BUCKET']}/#{value}"
```
cavis: Exchange (v3) uploads audio_files with a upload_path set to `path/to/file.mp3`.  But CMS (v4) uses the full `s3://bucket/path/to/file.mp3`.  Was getting some `no method 'starts_with?' for nil` error below, since the parsed value has no scheme.

This adds the mediajoint production s3 bucket prefix to those relative paths.

app/models/address.rb
```diff
@@ -1,7 +1,7 @@
 # encoding: utf-8
 
 class Address < BaseModel
-  belongs_to :addressable, polymorphic: true, touch: true
+  belongs_to :addressable, -> { with_deleted }, polymorphic: true, touch: true
```
cavis: Was seeing some errors where an account/user had been deleted `https://cms.prx.org/api/v1/accounts/12482`

... but the address representer is unable to render `https://cms.prx.org/api/v1/stories/12132`, since from its side `address.account` is nil.

svevang: This looks good.  I see that on the story, we are doing the exact same thing (using the `with_deleted` scope).

In this particular case, story 12132 is missing it's `creator_id` (it's nil), so this change wont help with that particular 500.

app/models/concerns/fixerable.rb
svevang: Looks good, maybe worth a comment?

cavis: Yeah, I'll make this more explicit.

app/models/address.rb
cavis: Hrmm... what is that nil creator_id going to break?  I don't see the `:creator` being represented anywhere.

svevang: Sorry, that `creator_id` is a red herring.

I have your branch checked out and looking at that `stories/12132` endpoint (using my dump of staging data), I see: 
```
From: /app/app/representers/api/address_representer.rb @ line 15 Api::AddressRepresenter#self_url:

    13: def self_url(address)
    14:   binding.pry
 => 15:   api_account_address_path(address.account)
    16: end

[1] pry(#<Api::AddressRepresenter>)> address.addressable.accounts
D, [2018-08-22T16:31:39.384522 #18] DEBUG -- :   User Load (0.6ms)  SELECT  `users`.* FROM `users` WHERE `users`.`id` = 12482 LIMIT 1
D, [2018-08-22T16:31:39.403896 #18] DEBUG -- :   Account Load (0.5ms)  SELECT `accounts`.* FROM `accounts` INNER JOIN `memberships` ON `accounts`.`id` = `memberships`.`account_id` WHERE `memberships`.`user_id` = 12482
=> []
[2] pry(#<Api::AddressRepresenter>)> address.addressable.present?
=> true
[3] pry(#<Api::AddressRepresenter>)> address.account.nil?
D, [2018-08-22T16:32:02.136454 #18] DEBUG -- :   Account Load (1.0ms)  SELECT  `accounts`.* FROM `accounts` INNER JOIN `memberships` ON `accounts`.`id` = `memberships`.`account_id` WHERE `memberships`.`user_id` = 12482 AND (type = 'IndividualAccount')  ORDER BY `accounts`.`id` ASC LIMIT 1
=> true
```

You can see the `SELECT ... FROM users` is using the non-paranoia scoped query, which is good!  However the user is missing his account (which all users should have, right?).

So in addition to removing the with_deleted scope on addressable, perhaps we should guard against users with no accounts in the representer?

Issue #437, opened 2018-08-07
UI for podcast import


svevang:
Provide an internal user interface for importing podcasts from an RSS.
Leverage existing PodcastImport/EpisodeImport models and the associated controller endpoints.

- [x] Retrieve bearer token from id service
- [x] Post RSS url to `podcasts_import_controller#create` 
- [x] Stub out interface on publish
- [x] ensure authorize access
- [x] ng testing
- [x] create import via Series post
- [x] test RSS URL to make sure that it's valid
- [ ] spinner and other indicators for episode import status
- [ ] incorporate style suggestions and designs
- [ ] guid validation
- [ ] import status tab is slow for large imports: needs pagination, or eliminate n+1 query in representer

PR #439, opened 2018-08-21
test relation presence before serializing

svevang:
Discovered this in pursuit of https://github.com/PRX/cms.prx.org/issues/437

This PR follows the series relation to ensure that if a user deletes a series, the podcast import representer will still render.


svevang:
Closing in favor of a more comprehensive roundup of podcast imports tweaks/fixes.

PR #440, opened 2018-08-22
Feat/prepare for podcast import ui

svevang:
In support of #437, this principally:

* creates a facility in the Series api resource for creating podcast imports
* modifies the podcast import process to work in conjunction with the UI


app/models/podcast_import.rb
```diff
@@ -13,89 +13,160 @@ class PodcastImport < BaseModel
...
   def create_or_update_episode_imports!
-    feed.entries.map do |entry|
-      entry_hash = feed_entry_to_hash(entry)
-      audio_files = entry_audio_files(entry_hash)
-      get_or_create_template(audio_files, entry_hash['enclosure'].type)
-      episode_import = create_or_update_episode_import!(entry_hash, audio_files)
-      episode_import.import_later
+    feed_entries, entries_with_dupe_guids = parse_feed_entries_for_dupe_guids
+
+    update_attributes(feed_episode_count: feed_entries.length)
```
cavis: Wait, so then `feed_episode_count` _won't_ be the number of `<item>`s in the RSS?  Seems like this is going to un-count any items with duplicate guids.

app/models/podcast_import.rb
```diff
@@ -13,89 +13,160 @@ class PodcastImport < BaseModel
...
+
+    created_imports
+  end
+
+  def enqueue_episode_import_jobs(created_imports)
+    messages = created_imports.map do |ei|
+      job = EpisodeImportJob.new(ei)
+      msg = {}
+      msg[:message_body] = job.serialize
+      msg[:message_attributes] = {
```
cavis: So this is manually queuing messages for shoryuken?  There's no built in way to schedule those?

app/models/podcast_import.rb
```diff
@@ -197,13 +274,19 @@ def update_image(purpose, image_url)
       to_insert << series.images.build(upload: clean_string(image_url), purpose: purpose)
     end
 
-    story.images.destroy(to_destroy) if to_destroy.size > 0
```
cavis: No longer need to destroy these?

app/representers/api/podcast_import_representer.rb
```diff
@@ -8,9 +8,16 @@ class Api::PodcastImportRepresenter < Api::BaseRepresenter
   property :status, writeable: false
   property :created_at, writeable: false
   property :updated_at, writeable: false
+  property :updated_at, writeable: false
```
cavis: 2x updated_at?

app/representers/api/podcast_import_representer.rb
```diff
@@ -8,9 +8,16 @@ class Api::PodcastImportRepresenter < Api::BaseRepresenter
...
+  property :feed, writeable: false
```
cavis: Is this a string?  (Guess i was just expecting something more like `feed_url`)

app/representers/api/series_representer.rb
```diff
@@ -9,6 +9,7 @@ class Api::SeriesRepresenter < Api::BaseRepresenter
   property :created_at, writeable: false
   property :updated_at, writeable: false
   property :app_version, writeable: false
+  property :import_url
```
cavis: I wonder if there's any case for only exposing the original-feed-url in the auth-series-representer

app/controllers/api/auth/podcast_imports_controller.rb
```diff
@@ -27,11 +27,38 @@ def create_resource
...
+    authorization.podcast_imports
+  end
+
+  def verify_rss
+    verify_rss_resource.tap do |podcast_import|
+
+      status = begin
+                 podcast_import.get_feed
+                 :ok
+               rescue StandardError
```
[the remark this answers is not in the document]
svevang: Doing a little bit of message passing here with exceptions:
https://github.com/PRX/cms.prx.org/blob/master/app/models/podcast_import.rb#L129

I think an alternative would be to track validations as a set of validation keys and messages (or even a single message) and then pass that on through the representer back to the client.

app/controllers/api/series_controller.rb
```diff
@@ -47,6 +47,26 @@ def create_resource
...
 
+  # XXX this method is copied from the ResourceCallbacks concern
+  def create
+    json_body_params = JSON.parse(request.body)
+
+    series = create_resource
+    consume! series, create_options
+    hal_authorize series
+
+    @series = if json_body_params.has_key?('import_url')
```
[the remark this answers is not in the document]
svevang: It's possible. I think the problem lies in the `create_from_feed` call below.  As I understand it, roar/hal_api_rails must create model instances in a rails conventional manner, so a custom method to create the series from a feed mucks that up.

Another tact to clean this up would be to expose these params via an internal API (e.g. `hal_params.permit(:import_url)`.  But these params are not exposed anywhere (?), which in hindsight appears to be a conscious decision in the hal_api/roar/representer world.

Assuming we stick with something like what's currently here, these params likely need to be wrapped in strong params to prevent nested attr hacks.

app/jobs/episode_import_job.rb
```diff
@@ -1,7 +1,7 @@
 # encoding: utf-8
 
 class EpisodeImportJob < ApplicationJob
-  queue_as :cms_default
+  queue_as :cms_podcast_import
```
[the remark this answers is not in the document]
svevang: Yep, it really hammers things, so hopefully this will keep the queue generally moving.

app/models/concerns/searchable.rb
```diff
@@ -6,22 +6,36 @@ module Searchable
   included do
     include Elasticsearch::Model
 
+    attr_accessor :skip_searchable
+
     after_commit on: [:create, :update] do
-      if Rails.env.test?
```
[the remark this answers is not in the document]
svevang: I don't think it makes a difference as `after_commit` hooks don't run in tests, or at least I have had to manually invoke the after_commit hooks inside tests.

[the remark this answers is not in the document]
svevang: All the changes in here are really ugly, and most of the structure here is so I could get a test to cover this. ☹️

app/models/podcast_import.rb
svevang: Yep, that's a mistake! Thanks for catching.  When I renamed that variable, I shifted the semantics, so I need to adjust my arithmetic here and over in publish:
https://github.com/PRX/publish.prx.org/pull/538/files#diff-92016921774e9e471370aba107bc8249R62

app/models/podcast_import.rb
svevang: So, yeah. I'm not super happy about this, but this change shaved about 1-2 minutes off the podcast import run (which now runs in about 20 seconds for n=1000 episodes).

The story is that enqueuing jobs on SQS is _really_ slow. There is a bulk api for shoryuken workers I think.  Maybe I should have ported the podcast import job to become a shoryuken worker instead?

app/models/podcast_import.rb
svevang: There is no story attribute here. I looked around and bisected,  Is it vestigial?

app/representers/api/podcast_import_representer.rb
svevang: This is the output of Feedjira: a nested JSON object containing the parsed structure of the podcast rss.  There was some oscillation in design between how to validate the rss (client versus server). So in one timeline, this was used on the client side to display things like images.

Now it acts more as a sentinel and sort of maps to the semantics of the `verify-rss` endpoint: "hay validate this podcast rss and give me the validated data." Could be useful?

app/representers/api/series_representer.rb
svevang: Hmm, not sure I understand. `original-feed-url` as in "this feed url is where this series was derived from"?

The representer here really bears the burden of the underlying resources.

app/controllers/api/auth/podcast_imports_controller.rb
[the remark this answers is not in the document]
cavis: Ah, gotcha.  Those methods just raise generic StandardErrors.

app/controllers/api/series_controller.rb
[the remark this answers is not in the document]
cavis: Ah, I see.  So you'd need to somehow get the `current_user` and `account` at the same time.  And even then ... complicated.  If there has to be some "magic" somewhere, I'd prefer it here in the controller.

Does this properly belong here in the non-auth series controller?

app/models/podcast_import.rb
cavis: That is quite a bit quicker!  I guess my main objection is knowing the arbitrary structure of the messages here.

I guess I'm fine with leaving this in, as long as it's marked as TODO/MAGIC or something.  And maybe a ticket to come back and do a refactor at some point.

app/models/podcast_import.rb
cavis: Oh, interesting.  How did this ever work then?  I guess `to_destroy` has always been empty.

app/representers/api/podcast_import_representer.rb
cavis: Gotcha.

app/representers/api/series_representer.rb
cavis: Just wondering if there's a security issue with telling the world where our feeds originally came from.  Then again, feeds are public so why would we care?  Wondering withdrawn!

And yeah, if this moved to the auth representer, you couldn't set it via the public controller.

app/controllers/api/series_controller.rb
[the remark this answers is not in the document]
svevang: Re: `Does this properly belong here in the non-auth series controller?`

My instinct is to say no. But, there is some protection here, up on line 42 we check for the `current_user` which is enough to reject anonymous access (with a 500--access attr on nil).

It looks like we are missing a corresponding set of account scoped series routes within the api auth scope.  If we create those new routes (at least for the create method here), we'd get proper 4xx semantics. Worth it?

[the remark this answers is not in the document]
cavis: Yeah, let's track that separately.  In theory, those policy classes should be restricting write access to resources, and returning a 4xx long before it gets there.  May just be an edge case with the use of current_user.

So we should hash this out at some point ... but as we get more into this with ID creating accounts, my feelings are that it's fine to PUT/POST/DELETE to public api endpoints.  The `/api/v1/authorizations` routes were created to solve (1) caching issues on GET, and (2) filtering resources based on an auth token.  And neither of those are a consideration if you're creating/updating/deleting things.

app/models/concerns/import_utils.rb
```diff
@@ -13,14 +13,26 @@ module ImportUtils
...
   def default_story_url(story)
     StoryDistributions::EpisodeDistribution.default_story_url(story)
   end
 
+  def clean_title(str)
+    str = clean_string(str)
+    return str if str.length <= 255
+    truncated = str[0..254]
+
+    e = RuntimeError.new("ImportUtils: String length of #{str.length} exceeds title column length of 255: #{truncated}")
```
cavis: Does this include the story id?  Seems like that might be helpful for debugging later.

cavis:
:+allthenumbers:

PR #445, opened 2018-08-28
Elasticsearch authorization filters

pkarman:
Implements Stories authenticated API filters for authorization based on network or account. These `:search` rules should match the existing rules for the `:index` endpoint, and so the tests have been copied and adapted for the `:search` method.

The following URL parameters are supported for filtering: `:series_id`, `:account_id`, `:network_id` and `:app_version`. The `:app_version` is similar to `:v3` and `:v4` for the `:index` endpoint, and expects values like those (e.g. `app_version=v4`).

To support filtering by NULL values, the special string `NULL` is supported. So you can search for Stories that have an empty description: `fq[description]=NULL` or where the series_id is unassigned `series_id=NULL`.

Note the `fq` param is for "fielded queries" where you want to structurally limit a search by a specific field. You can also do that directly in the `q` (query) param: `q=foo:bar` and `fq[foo]=bar` should search the same way.

Sorting uses the `sort` param. (I think it is plural "sorts" on the `:index` endpoint but ES is singular so I left it singular for that consistency. It can be consistent for our API or for ES but not both.) To get the affect of SQL NULL sorting, special mangling of the sort direction is done to reflect NULL values.

### DIFF
--- a/app/controllers/api/auth/stories_controller.rb
+++ b/app/controllers/api/auth/stories_controller.rb
@@ -2,6 +2,7 @@
 
 class Api::Auth::StoriesController < Api::StoriesController
   include ApiAuthenticated
+  include ApiSearchable
 
   api_versions :v1
 
@@ -51,5 +52,25 @@ class Api::Auth::StoriesController < Api::StoriesController
     else
       authorization.token_auth_stories
     end
+  end
+
+  def search
+    @stories = Story.text_search(search_query, search_params, authorization)
+    index
+  end
+
+  def search_params
+    sparams = super
+    sparams['fq'] ||= {}
+    [:network_id, :series_id].each do |p|
+      if params[p]
+        sparams['fq'][p.to_s] = params[p]
+      end
+    end
+    sparams
+  end
+
+  def searchable_fields
+    Story.searchable_fields
   end
 end
--- a/app/controllers/api/auth/stories_controller.rb
+++ b/app/controllers/api/auth/stories_controller.rb
@@ -46,7 +46,8 @@ class Api::Auth::StoriesController < Api::StoriesController
   end
 
   def resources_base
-    # If there is a network_id specified, use that network
+    # If there is a network_id specified, use that network.
+    # The authz is performed in check_user_network().
     @stories ||= if params[:network_id]
       super.published
     else
@@ -55,7 +56,12 @@ class Api::Auth::StoriesController < Api::StoriesController
   end
 
   def search
-    @stories = Story.text_search(search_query, search_params, authorization)
+    # same logic as resources_base above
+    @stories ||= if params[:network_id]
+      Story.text_search(search_query, search_params)
+    else
+      Story.text_search(search_query, search_params, authorization)
+    end
     index
   end
 
--- a/app/controllers/api/auth/stories_controller.rb
+++ b/app/controllers/api/auth/stories_controller.rb
@@ -67,10 +67,10 @@ class Api::Auth::StoriesController < Api::StoriesController
 
   def search_params
     sparams = super
-    sparams['fq'] ||= {}
-    [:network_id, :series_id, :account_id].each do |p|
+    sparams[:fq] ||= {}
+    [:network_id, :series_id, :account_id, :app_version].each do |p|
       if params[p]
-        sparams['fq'][p.to_s] = params[p]
+        sparams[:fq][p.to_s] = params[p]
       end
     end
     sparams
--- a/app/controllers/api/auth/stories_controller.rb
+++ b/app/controllers/api/auth/stories_controller.rb
@@ -62,7 +62,7 @@ class Api::Auth::StoriesController < Api::StoriesController
   def search_params
     sparams = super
     sparams['fq'] ||= {}
-    [:network_id, :series_id].each do |p|
+    [:network_id, :series_id, :account_id].each do |p|
       if params[p]
         sparams['fq'][p.to_s] = params[p]
       end
--- a/app/controllers/concerns/api_searchable.rb
+++ b/app/controllers/concerns/api_searchable.rb
@@ -1,0 +1,19 @@
+# encoding: utf-8
+
+require 'active_support/concern'
+
+module ApiSearchable
+  extend ActiveSupport::Concern
+
+  def search_query
+    params.permit(:q)[:q]
+  end
+
+  def search_params
+    params.permit(:page, :size, :sort, fields: {}, fq: searchable_fields).to_h
+  end
+
+  def searchable_fields
+    []
+  end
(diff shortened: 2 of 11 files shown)


config/routes.rb
```diff
@@ -75,7 +75,9 @@
         resources :audio_files, except: [:new, :edit]
 
         resources :accounts, only: [:index, :show], module: :auth do
-          resources :stories, only: [:index, :create, :update]
+          resources :stories, only: [:index, :create, :update] do
+            get 'search', on: :collection
```
cavis: This got me thinking about pros/cons of having separate search endpoints vs query params.  But I think we already settled that in slack.  Plus we're not really stuck with anything until we publicize the `/search` endpoints in the links.

test/controllers/api/auth/stories_controller_test.rb
```diff
@@ -1,25 +1,28 @@
...
+  let (:v3_story) { create(:story_v3, account: account, description: 'v3_story') }
   let (:released_story) { create(:story, account: account, released_at: Time.now + 1.day) }
 
   before do
     account.stories.each { |s| s }
-    unpublished_story.update!(published_at: nil)
-    published_story.update_attributes!(published_at: 2.days.ago)
-    latest_story.update_attributes!(published_at: 1.day.ago)
-    released_story.update!(published_at: nil)
+    network.stories.each { |s| s.update!(short_description:"network #{s.id}") }
```
cavis: Can these short descriptions not be set via the factory `create()` calls?  Seems like the other things in here are trying to get around the un-set-ability of `published_at` (which could probably also be addressed in the factory).

test/controllers/api/auth/stories_controller_test.rb
```diff
@@ -98,6 +101,91 @@
...
+    describe 'search' do
+      before do
+        ElasticsearchHelper.new.create_es_index(Story)
+      end
+
+      def stories
+        assigns[:stories].to_a
+      end
+
+      it 'searches stories under their account' do
```
cavis: +1 to duplicating the non-search tests!

app/controllers/api/auth/stories_controller.rb
```diff
@@ -45,11 +46,37 @@ def filtered(resources)
...
     @stories ||= if params[:network_id]
       super.published
     else
       authorization.token_auth_stories
     end
   end
+
+  def search
+    # same logic as resources_base above
+    @stories ||= if params[:network_id]
```
cavis: Something is slightly wonky when searching with a network_id.  Digging around in the SQL a bit, the difference seems to be the `Story.network_visible` scope is applied by `api/v1/networks/123/stories` but not `api/v1/stories/search?q=network_id:123`.

app/controllers/api/auth/stories_controller.rb
```diff
@@ -45,11 +46,37 @@ def filtered(resources)
...
+    else
+      Story.text_search(search_query, search_params, authorization)
+    end
+    index
+  end
+
+  def search_params
+    sparams = super
+    sparams[:fq] ||= {}
+    [:network_id, :series_id, :account_id, :app_version].each do |p|
```
cavis: Just to note ... it's different that these are all specified via the `?q` param instead of coming from the route or the `?v4` filter.  Not bad, just different.

Also a bit awkward that the JSON representers render camelCase, but the search params (and a lot of our current query filters/sorts/etc) are under_scored.  But that's a bigger problem for another day.

app/controllers/api/auth/stories_controller.rb
cavis: Another "funny" discrepancy ... `api/v1/accounts/123/stories` is actually the equivalent of `account.public_stories`, not `account.stories.published`.  The `public_stories` includes some networks logic as well.

And now that i'm searching for that, it appears that networks/series/accounts filters all use "public" instead of "published".

test/controllers/api/auth/stories_controller_test.rb
pkarman: I was setting the descriptions just to make it easier to debug while testing. I'm happy to remove them now if that's cleaner.

cavis: Oh, i see, they're not really used in the tests themselves.  Yeah, then remove away.

app/controllers/concerns/api_searchable.rb
```diff
@@ -0,0 +1,19 @@
...
+
+module ApiSearchable
+  extend ActiveSupport::Concern
+
+  def search_query
+    params.permit(:q)[:q]
+  end
+
+  def search_params
+    params.permit(:page, :size, :sort, fields: {}, fq: searchable_fields).to_h
```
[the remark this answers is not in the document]
pkarman: That would allow you to specify a subset of fields that should be searched by default. It's not used right now, and in fact poses a security hole, so I've just removed it in b8f209d

app/controllers/api/auth/stories_controller.rb
pkarman: I think they are coming from the Rails `params` which are populated from the URL routing. So not from the `q` param. Unless I am misunderstanding?

cavis: Ah, I see.  So someday when we alter the routes to include sub-searches (like `/api/v1/series/123/stories`), those will start working.  But with this PR, the only way to use them is via the `?q=series_id:123` param.

pkarman: I see these routes defined (and tested).
```
                                   search_api_stories GET     /api/:api_version/stories/search(.:format)                                                              api/stories#search {:api_version=>"v1"}
             search_api_authorization_account_stories GET     /api/:api_version/authorization/accounts/:account_id/stories/search(.:format)                           api/auth/stories#search {:api_version=>"v1"}
              search_api_authorization_series_stories GET     /api/:api_version/authorization/series/:series_id/stories/search(.:format)                              api/auth/stories#search {:api_version=>"v1"}
                     search_api_authorization_stories GET     /api/:api_version/authorization/stories/search(.:format)                                                api/auth/stories#search {:api_version=>"v1"}
             search_api_authorization_network_stories GET     /api/:api_version/authorization/networks/:network_id/stories/search(.:format)                           api/auth/stories#search {:api_version=>"v1"}
```

cavis: Whoops!  I must have just missed those.

test/controllers/api/auth/stories_controller_test.rb
pkarman: I removed most of the descriptions, but left in a couple just as a way to explicitly get the factory objects indexed in the `begin` block.

Issue #443, opened 2018-08-24
ES queue incremental re-indexing


pkarman:
Use shoryuken queues to re-index documents on create/update/delete.

cavis:
May also want to consider a say_when job to do full reindexing every night.

PR #450, opened 2018-09-06
Search indexing

cavis:
For #443.

- [x] Use shoryuken to index series/story docs in the background
- [x] Use say_when to run a full reindex every night
  - NOTE: i used the already-boilerplate'd methods to do this.  Which happen to do the index-then-replace method.  So any changes that happen _during_ the minutes the indexer is running will be lost, until the next night's full reindex.

### DIFF
--- a/app/jobs/search_deindexer_job.rb
+++ b/app/jobs/search_deindexer_job.rb
@@ -1,0 +1,10 @@
+# encoding: utf-8
+
+class SearchDeindexerJob < ApplicationJob
+  queue_as :cms_search_indexer
+
+  def perform(class_name, id)
+    klass = class_name.constantize
+    klass.new(id: id).remove_from_index
+  end
+end
--- a/app/models/concerns/searchable.rb
+++ b/app/models/concerns/searchable.rb
@@ -39,7 +39,7 @@ module Searchable
     end
 
     # class method shorthand, useful for triggering via rake task or console or tests
-    def self.rebuild_index
+    def self.rebuild_index(opts = {})
       stager = index_stager
       indexer(stager).run
       stager.alias_stage_to_tmp_index && stager.promote
--- a/app/models/concerns/searchable.rb
+++ b/app/models/concerns/searchable.rb
@@ -30,7 +30,7 @@ module Searchable
     end
 
     def remove_from_index
-      __elasticsearch__.destroy_document
+      __elasticsearch__.delete_document
       refresh_index
     end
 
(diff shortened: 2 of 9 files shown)


app/jobs/search_deindexer_job.rb
```diff
@@ -0,0 +1,10 @@
+# encoding: utf-8
+
+class SearchDeindexerJob < ApplicationJob
+  queue_as :cms_search_indexer
+
+  def perform(class_name, id)
```
cavis: This may depend on the model, and if it's a real or soft-delete.  But generally ... by the time this job gets run, the ActiveRecord model isn't accessible.  (Also that global-id stuff does a `ModelClass.find(id)`, which isn't going to include `with_deleted` anyways).

app/models/concerns/searchable.rb
```diff
@@ -39,7 +39,7 @@ def refresh_index
     end
 
     # class method shorthand, useful for triggering via rake task or console or tests
-    def self.rebuild_index
+    def self.rebuild_index(_opts = {})
```
cavis: The `say_when` cron jobs always pass an opts object.  Whether you want it or not.

config/initializers/say_when.rb
```diff
@@ -27,6 +27,32 @@
...
+rescue ActiveRecord::StatementInvalid => ex
+  puts "Failed to init say_when job: #{ex.inspect}"
+end
+
+begin
+  SayWhen.schedule(
+    group: 'application',
+    name: 'reindex_series',
+    trigger_strategy: 'cron',
+    trigger_options: { expression: '0 0 7 * * ? *', time_zone: 'UTC' },
```
cavis: (It uses a slightly different "extended" cron syntax)

cavis:
It's also a bit fun when 2 people are developing at once ... getting the same messages off the dev SQS.

test/jobs/search_deindexer_job_test.rb
```diff
@@ -0,0 +1,27 @@
+require 'test_helper'
+
+class MockDeindexerModel
```
pstjean: I like the idea of just declaring these mocks inline. Reading these tests is slowly helping me get over withdrawal from RSpec

app/jobs/search_deindexer_job.rb
pstjean: Metaprogramming++

PR #452, opened 2018-09-10
Fix reindexing

cavis:
Full reindexes were failing because of a problem in the [elasticsearch-indexstager](https://github.com/18F/elasticsearch-indexstager-gem/blob/master/lib/elasticsearch/index_stager.rb#L42) gem.  Since that repo is archived, and we're not actually doing parallel reindexing ... I opted to just remove it for now.

Also preloads the story+series together, to make that index rebuild way faster.

### DIFF
--- a/app/models/concerns/searchable.rb
+++ b/app/models/concerns/searchable.rb
@@ -23,9 +23,9 @@ module Searchable
     end
 
     # shorthand methods
-    def reindex
+    def reindex(refresh = false)
       __elasticsearch__.index_document
-      refresh_index
+      refresh_index if refresh
       self
     end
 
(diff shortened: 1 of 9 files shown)


app/models/concerns/searchable.rb
```diff
@@ -23,9 +23,9 @@ module Searchable
     end
 
     # shorthand methods
-    def reindex
+    def reindex(refresh = false)
       __elasticsearch__.index_document
-      refresh_index
+      refresh_index if refresh
```
cavis: Also decided not to refresh the index on create/update.  But it does refresh on destroy.  Hoping that will speed up those operations a little bit.  (They're async processed anyways, so user really shouldn't notice).

PR #454, opened 2018-09-17
Surface worker errors in logs

svevang:
This PR has the goal of reducing/eliminating exceptions within the shoryuken worker machinery. Assuming that we are allowing long-running jobs (20-30 seconds) on our worker queues, then for big job batches, it's possible to consume all of our database connections. Assuming the db pool < # worker and a short db pool request timeout.

First, this PR contains a minor tweak to surface exceptions in the worker log. Looking something like:
```
ERROR: Process died, reason: Error while trying to deserialize arguments: could not obtain a database connection within 5.000 seconds (waited 5.001 seconds)
```
This should allow us to get a handle on how many workers are dying.

Secondly, we set up a env/secrets change for the workers: increase the `DATABASE_POOL_SIZE` to match `CONCURRENCY` (per the recommendations of shoryuken) plus the database needs of `say_when`.

### DIFF
--- a/app/jobs/application_job.rb
+++ b/app/jobs/application_job.rb
@@ -4,6 +4,7 @@ class ApplicationJob < ActiveJob::Base
 
   rescue_from(StandardError) do |e|
     NewRelic::Agent.notice_error(e)
+    raise e
   end
 
   around_perform do |job, block|


app/jobs/application_job.rb
```diff
@@ -4,6 +4,7 @@ class ApplicationJob < ActiveJob::Base
 
   rescue_from(StandardError) do |e|
     NewRelic::Agent.notice_error(e)
+    raise e
```
cavis: Wouldn't we already be getting a count of these in NewRelic?

svevang: Maybe.

Trying to verify: Searching in cloudwatch (in the production cms stack log group) for "Could not log", I'm seeing the last worker error that I think is associated with this PR is from 9-9-18, which is off the end of the NR data.

New Relic also needs to be upgraded to capture worker data.  Clicking on the worker/background tab, there's no data and a message to upgrade the account: "To start using Background Jobs, upgrade your subscription level.".

cavis: Yeah, we don't use a paid NR plan for our ECS stuff.  Mainly due to cost.  But we could revisit that.  But the `notice_error` should still show up NR, even if it doesn't know that's a "worker".

So one impact of this (and I'm seeing [feeder has the same problem](https://github.com/PRX/feeder.prx.org/blob/master/app/jobs/application_job.rb#L6)) is that it will change the retry logic.  Re-throwing will start using the SQS redrive policy - delay 1 minute, retry up to 10 times, then DLQ it.

But maybe that's a good thing?  We've certainly gotten into trouble by not retrying things like db connection timeouts.

svevang: It's going to be easier to track if we can surface these errors in logs etc, and be able to inspect the DLQ to see which jobs are stuck. We may need to preserve the existing queue behavior, in which case we should adjust `maxReceiveCount` to 1, so that failed jobs are immediately deposited in the LDQ queue:
https://docs.aws.amazon.com/AWSSimpleQueueService/latest/SQSDeveloperGuide/sqs-dead-letter-queues.html
Depends on if we should retry stuck jobs?

svevang: For CMS, the `EpisodeImport` and `PodcastImport` are both idempotent, not so sure about the search jobs.

cavis: Yeah, indexer jobs are fine to retry.  (Though less catastrophic if they just fail).

PR #464, opened 2018-10-11
independently modify the worker's database pool size

svevang:
This PR sets up a new way to tune the database pool size for the CMS workers. By setting a new `WORKER_DATABASE_POOL_SIZE` env var, we can override the default `DATABASE_POOL_SIZE`, but only in the shoryuken worker pool--not the web process.  This will help with database timeouts for long running concurrent access to the database across workers.

I've been setting `WORKER_DATABASE_POOL_SIZE` as `WORKER_COUNT  + 5` to account for the Shoryuken workers plus the SayWhen pollers.

This will require an update to the aws-secrets.


env-example
```diff
@@ -23,3 +23,5 @@ PUBLIC_ASSET_SECRET=
 RAILS_ENV=development
 RAILS_SECRET_KEY=seeecrets
 ELASTICSEARCH_URL=http://elasticsearch:9200
+WORKER_DATABASE_POOL_SIZE=15
+WORKER_COUNT=10
```
cavis: Is that just a good default for dev?

PR #471, opened 2018-10-19
Account creation

pstjean:
Allow account creation, via the authorized account endpoint as well as through the authorized user endpoint

### DIFF
--- a/app/controllers/api/accounts_controller.rb
+++ b/app/controllers/api/accounts_controller.rb
@@ -5,6 +5,15 @@ class Api::AccountsController < Api::BaseController
 
   represent_with Api::AccountRepresenter
 
+  # filter_resources_by :user_id
+
+  def create_resource
+    return super unless user
+
+    user.create_individual_account
+    user.individual_account
+  end
+
   private
 
   def resources
--- a/app/controllers/api/accounts_controller.rb
+++ b/app/controllers/api/accounts_controller.rb
@@ -12,7 +12,7 @@ class Api::AccountsController < Api::BaseController
   end
 
   def resources_base
-    user.try(:accounts) || Account
+    user.try(:accounts) || super
   end
 
   def user
--- a/app/controllers/api/auth/accounts_controller.rb
+++ b/app/controllers/api/auth/accounts_controller.rb
@@ -12,11 +12,10 @@ class Api::Auth::AccountsController < Api::AccountsController
   before_filter :ensure_write_scope, except: [:index, :show]
 
   def create_resource
-    return super unless params['user_id']
+    return super unless user
 
-    account_user = User.find(params['user_id'])
-    account_user.create_individual_account
-    account_user.individual_account
+    user.create_individual_account
+    user.individual_account
   end
 
   private
--- a/app/controllers/api/auth/accounts_controller.rb
+++ b/app/controllers/api/auth/accounts_controller.rb
@@ -7,15 +7,6 @@ class Api::Auth::AccountsController < Api::AccountsController
 
   represent_with Api::Auth::AccountRepresenter
 
-  filter_resources_by :user_id
-
-  def create_resource
-    return super unless user
-
-    user.create_individual_account
-    user.individual_account
-  end
-
   private
 
   def resources
--- a/app/controllers/api/auth/accounts_controller.rb
+++ b/app/controllers/api/auth/accounts_controller.rb
@@ -7,6 +7,13 @@ class Api::Auth::AccountsController < Api::AccountsController
 
   represent_with Api::Auth::AccountRepresenter
 
+  filter_resources_by :user_id
+
+  def after_create_resource(res)
+    account_user = User.find(params['user_id'])
+    account_user.individual_account = res
+  end
+
   private
 
   def resources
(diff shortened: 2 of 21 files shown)


app/models/account.rb
```diff
@@ -44,6 +44,9 @@ def self.policy_class
   end
 
   private
+  def individual_acct?
```
pstjean: To me, it's an open question whether this is the way this should be handled. Because `IndividualAccount` overrides the `path` getter to use the opener's login, and because we essentially duplicate these validations on login for the user, _and_ because creating an individual account is effected by using `User#create_individual_account` which copies `login` into `Account.path` there is _currently_ little value in validating both.

On the other hand, things could drift in the future. Maybe writing custom validations using `self[:path]` is the way to go?

app/controllers/api/accounts_controller.rb
```diff
@@ -12,7 +12,7 @@ def resources
   end
 
   def resources_base
-    user.try(:accounts) || Account
+    user.try(:accounts) || super
```
pstjean: This was an extant bug. We always want this method to return an `ActiveRecord::Relation`.

app/controllers/api/auth/accounts_controller.rb
```diff
@@ -7,6 +7,15 @@ class Api::Auth::AccountsController < Api::AccountsController
 
   represent_with Api::Auth::AccountRepresenter
 
+  filter_resources_by :user_id
+
+  def create_resource
+    return super unless params['user_id']
```
pstjean: We want to override the default resource creation logic if the account is being created through the `/api/v1/auth/users/:user_id/account` endpoint.

app/controllers/api/auth/accounts_controller.rb
```diff
@@ -9,15 +9,22 @@ class Api::Auth::AccountsController < Api::AccountsController
...
+
     account_user = User.find(params['user_id'])
     account_user.create_individual_account
     account_user.individual_account
   end
 
   private
 
+  def ensure_write_scope
+    user_not_authorized unless prx_auth_token.scopes&.include? 'account:write'
```
pstjean: This is just a placeholder for now, scope currently doesn't exist. I'm using this colon separated deal for permissions on an entity which I've seen e.g. in the [Slack API](https://api.slack.com/docs/oauth-scopes).

config/routes.rb
```diff
@@ -77,7 +77,13 @@
       resource :authorization, only: [:show] do
         resources :audio_files, except: [:new, :edit]
 
-        resources :accounts, only: [:index, :show], module: :auth do
+        resources :users, except: [:new, :edit, :create, :destroy], module: :auth do
```
cavis: Hmm... so I'm a little hesitant to have an `auth/users` endpoint.  The point of the `auth/accounts` endpoint was that it _only_ returned accounts in your auth token.  It's a bit strange to have "all-users" here.

Trying to think what it might mean to POST to the `auth/accounts` endpoint.  I guess that _should_ create a new account (if you're allowed), and set you as the opener.  Then you could POST a new `auth/accounts/1234/memberships` with a `set_user_uri` or something.

app/models/user.rb
```diff
@@ -15,6 +15,8 @@ class User < BaseModel
 
   after_commit :create_individual_account, on: [:create]
 
+  validates_uniqueness_of :login
```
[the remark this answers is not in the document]
pstjean: Yeah this is just a vestige of some of my earlier fumbling around. I'll leave it, but it's not going to be needed for this feature.

config/initializers/route_reservations.rb
```diff
@@ -0,0 +1,11 @@
+ROUTE_RESERVED_WORDS = %w(v1 oauth_clients oauth me api s3_uploads
```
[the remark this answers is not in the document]
pstjean: Yeah, it hurt to copy and paste this. Drift is possible in the future.

app/models/account.rb
pstjean: Yes, there totally could. Requires some further thought 🤔

app/controllers/api/auth/accounts_controller.rb
```diff
@@ -7,8 +7,24 @@ class Api::Auth::AccountsController < Api::AccountsController
...
   represent_with Api::Auth::AccountRepresenter
 
+  filter_resources_by :user_id
+
+  before_filter :ensure_write_scope, except: [:index, :show]
+
+  def create_resource
+    return super unless params['user_id']
+
+    account_user = User.find(params['user_id'])
```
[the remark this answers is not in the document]
pstjean: So it does! Fixed.

app/models/account.rb
pstjean: Ok, wrote validations that work directly on the path attribute, and the API now returns a `409` with the failure message when those validations fail.

app/controllers/api/auth/accounts_controller.rb
```diff
@@ -7,8 +7,24 @@ class Api::Auth::AccountsController < Api::AccountsController
 
   represent_with Api::Auth::AccountRepresenter
 
+  filter_resources_by :user_id
+
+  before_filter :ensure_write_scope, except: [:index, :show]
```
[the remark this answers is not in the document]
pstjean: Hey that's a cool feature! Done.

config/routes.rb
pstjean: This is a good point. I've constrained the routes so that for now, the only thing that exists on `api/v1/authorization/users` is the `api/v1/authorization/users/:user_id/accounts` endpoint. If and when we expand account creation via CMS, that's worth another look.

PR #481, opened 2018-11-16
Feat/fixup cms auth 500

svevang:
Fixes up a 500 error when calling POST on the non-auth version of the `api_account_series#create`.  Everything underneath `api/<version>/authorization` is treated as non-cacheable, which is synonymous with writing to resources (and therefor being authenticated).

This PR allows for authenticating writable api resources that are *not* under the `/api/<version>/authorization` path.

### DIFF
--- a/app/controllers/api/base_controller.rb
+++ b/app/controllers/api/base_controller.rb
@@ -18,6 +18,8 @@ class Api::BaseController < ApplicationController
   include PolymorphicResource
   include AnnounceActions
 
+  before_filter :authenticate_user!, only: [:create, :update, :delete]
+
   protect_from_forgery with: :null_session
 
   allow_params :show, [:api_version, :format, :zoom]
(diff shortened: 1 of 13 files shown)


app/controllers/api/series_controller.rb
```diff
@@ -3,6 +3,8 @@
 class Api::SeriesController < Api::BaseController
   include ApiSearchable
 
+  before_filter :authenticate_user!, only: :create
```
cavis: Update/delete don't have the same problems?

Also wondering about the other public controllers (stories, etc).  What would happen if this before filter just moved to the base controller itself?  (Since we theoretically don't allow any non-auth'd POST/PUT/DELETE).

app/controllers/api/base_controller.rb
```diff
@@ -68,6 +68,10 @@ def authorization
     Authorization.new(prx_auth_token) if prx_auth_token
   end
 
+  def authenticate_user!
```
cavis: Makes sense.  Probably more sense than having this in the concern, since `current_user` is defined here.

PR #492, opened 2018-12-06
Account creation authentication

pstjean:
Real chicken and egg problem here: currently, all permissions are based on accounts. The `authorized_resources` in a token issued by ID are the IDs of accounts that the user is either a member or admin of.

So, what account do you need to belong to to create the very first account you’ll belong to 🤔

I've decided after a lot of back and forth to ensure that the token contains the admin scope for the PRX Administrator account. This means that the user the token was issued to has a membership in that account.

It would be great if the subject of the JWT was the user in question, but it seems that it's actually referring to the `user_id` column in `client_applications`


app/policies/account_policy.rb
```diff
@@ -1,6 +1,6 @@
 class AccountPolicy < ApplicationPolicy
   def create?
-    token.present? && token.scopes&.include?('account:write')
+    token && token.authorized?(ENV['PRX_ADMIN_ID'], :admin)
```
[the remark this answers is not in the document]
svevang: I'm assuming this `PRX_ADMIN_ID` is the magical number 8 _user_ id? Might be nice to see `PRX_ADMIN_USER_ID` in there just so there's no confusion.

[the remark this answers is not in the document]
svevang: Just to clarify, this means that regular (non-admin) users cannot create accounts using this policy-- just oauth authorized applications (like `id.prx.org`) will create accounts?

env-example
```diff
@@ -32,3 +32,4 @@ RAILS_SECRET_KEY=seeecrets
 ELASTICSEARCH_URL=http://elasticsearch:9200
 WORKER_DATABASE_POOL_SIZE=15
 WORKER_COUNT=10
+PRX_ADMIN_ID=
```
svevang: If this is the user id, would it make sense to default this to `8` here?

app/policies/account_policy.rb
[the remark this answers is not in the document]
cavis: Hmm... I'm not sure I like having any user_ids in CMS.  Or is this actually an account id?

So does the `account:write` scope not work?  I'd think a "normal" user/client wouldn't be able to ask for that.  ID would only issue that scope to itself, for specific purposes.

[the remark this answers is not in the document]
pstjean: Yeah, that's an account id. I don't like this either.

We could use scopes, to do that this ticket will require a fair amount of functionality added on top of the existing OAuth2 implementation. ID has no way to control which scopes it issues, beyond ensuring that they're on our whitelist. I think the proper way to implement this would be to add scopes as a property of client applications, and then grab them from the database to add them to the token.

I discussed this with Andrew and what we landed on is that since individual accounts can't be created without the ID of a user with no account, and all other account types require admin approval, this isn't worth adding substantial new functionality to our auth system. We just allow anyone with a valid token to create an individual account for a user who doesn't have one (i.e., the existing policy). Then, if we end up finding that there are other use cases for scopes that grant permissions, we can revisit.

pstjean:
Ok! Added DatabaseCleaner using the transaction strategy, so each test is now wrapped in a separate ActiveRecord transaction.  I was able to reproduce the test failures locally with a seed from CI, and this has resolved them.

We could probably go through the test suite now and remove all the more ad hoc efforts at data cleanup that have accrued over time. I removed one that I added earlier in this PR.

Doesn't seem to have impacted test suite time negatively:

```
Without DatabaseCleaner
-----------------------
Finished in 270.386185s
Finished in 259.705708s
Finished in 287.049458s

With DatabaseCleaner
-----------------------
Finished in 288.298893s
Finished in 269.072978s
Finished in 253.921726s
```

app/controllers/api/accounts_controller.rb
```diff
@@ -1,17 +1,22 @@
...
   # filter_resources_by :user_id
 
+  # Unless there's already an opener, set it to the authenticated user
   def create_resource
-    return super unless user
-
-    user.create_individual_account
-    user.individual_account
+    super.tap do |acct|
+      acct.opener_id ||= current_user.id
```
svevang: Should a user be able to create a GroupAccount/Account here via `kind` param? Or just an IndividualAccount?

Gemfile
```diff
@@ -102,6 +102,7 @@ end
 group :test do
   gem 'codeclimate-test-reporter', require: false
   gem 'codecov', require: false
+  gem 'database_cleaner'
```
svevang: Yes! This makes me irrationally exuberant.

Might need to add gemfile.lock changes?

app/controllers/api/accounts_controller.rb
pstjean: My understanding is that any user should be able to create a GroupAccount/StationAccount without restriction, since there's additional logic in place for administrators to approve those other account types. The policy that would handle those other account types is the existing account policy, which just checks for the presence of a token.

Gemfile
pstjean: Whoops! Yeah, those are now added. Don't know how I missed them.

app/policies/individual_account_policy.rb
```diff
@@ -0,0 +1,8 @@
+class IndividualAccountPolicy < AccountPolicy
+  def create?
+    # An account without a membership to an Individual Account
```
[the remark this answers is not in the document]
pstjean: Quiet, hound! No can know.

app/controllers/api/accounts_controller.rb
cavis: I don't _think_ GroupAccounts need admin approval?  But yeah, StationAccounts definitely do.  The only restriction is probably that the creating user has to have a confirmed email.  (But ID shouldn't be giving out tokens to unconfirmed users anyways).

PR #497, opened 2019-01-10
fixup 500s: don't render paths for missing resources

svevang:
This fix tests if the record is present in the database before rendering the associated link.


pstjean:
👍

Issue #470, opened 2018-10-19
No status set with 0 audio files


cavis:
Having 0 audio files doesn't trigger a `status / status_message` on the AudioVersion.  Not sure if this was intentional (to avoid an invalid status before you've even upload anything).  But the result is that CMS can publish a story with 0 audio_files.  (Publish _does_ prevent this, though).

PR #499, opened 2019-01-29
Validate > 0 audio files

cavis:
For #470.  Set a status/status_message on the story when there aren't any audio_files.

Main impact should be preventing accidental publishing of a story with 0 audio.  Not sure that's happened in real life, but can't be too careful.

### DIFF
--- a/app/models/story.rb
+++ b/app/models/story.rb
@@ -326,12 +326,12 @@ class Story < BaseModel
 
   def set_status
     av_errors = ''
-    if audio_versions.empty?
-      av_errors << "Story '#{title}' has no audio."
+    if audio_files.empty?
+      av_errors << "Story '#{title}' has no audio"
     else
       audio_versions.each do |av|
         if !av.compliant_with_template? || av.status == INVALID
-          av_errors << "Invalid audio version: '#{av.label}.' "
+          av_errors << "Invalid audio version: '#{av.label}' "
         end
       end
     end
--- a/test/factories/audio_version_factory.rb
+++ b/test/factories/audio_version_factory.rb
@@ -12,7 +12,7 @@ FactoryGirl.define do
       audio_files_count 1
     end
 
-    after(:create) do |audio_version, evaluator|
+    after(:create, :stub) do |audio_version, evaluator|
       c = evaluator.audio_files_count.to_i
       (1..c).each { |i| FactoryGirl.create(:audio_file, audio_version: audio_version, position: i) }
       audio_version.reload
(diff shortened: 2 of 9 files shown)


kookster:
How does this work for stories that have a video instead of an audio file, like that 99pi episode? @cavis

cavis:
We still attach those video files as a `story.audio_versions[].audio_files[]`.  So they would pass this validation.  This doesn't care what the file is ... the story just has to have some attached "audio_file" record.

app/models/story.rb
```diff
@@ -326,12 +326,12 @@ def update_published_to_released
 
   def set_status
     av_errors = ''
-    if audio_versions.empty?
-      av_errors << "Story '#{title}' has no audio."
+    if audio_files.empty?
```
cavis: After trying some different things, I realized the Story already checks for `audio_versions.empty?`.  From there, it's not very controversial to test for `audio_files.empty?`.

This has the added bonus of not breaking Publish.  Setting an `AudioVersion.status_message` would show up right away in the Publish UI (in the audio-uploader widget).  Whereas a `Story.status_message` only shows up if Publish can't find anything else wrong and you're about to publish a story.

test/factories/audio_version_factory.rb
```diff
@@ -12,10 +12,11 @@
       audio_files_count 1
     end
 
-    after(:create) do |audio_version, evaluator|
+    after(:create, :stub) do |audio_version, evaluator|
       c = evaluator.audio_files_count.to_i
       (1..c).each { |i| FactoryGirl.create(:audio_file, audio_version: audio_version, position: i) }
       audio_version.reload
+      audio_version.length(true)
```
cavis: Reloading the cached length fixed validations for a bunch of tests, thus breaking them.

test/factories/audio_version_template_factory.rb
```diff
@@ -1,10 +1,11 @@
 FactoryGirl.define do
   factory :audio_version_template do
-    series
```
cavis: This was causing a bit of recursion, since the series is going to create an audio version, which will create more audio version templates.  Some tests ran way faster after removing.

Makes me wonder what other factory recursion we may have going on in the codebase.

app/models/story.rb
svevang: +1 So basically this is saying, no matter what configuration of audio versions, we expect at least one audio file in a publishable story

cavis: Exactly.  It seems like this was already a business requirement, and enforced with _most_ audio versions (those with a "duration > 0" validation).  But we just weren't catching this edge case.

PR #507, opened 2019-03-07
Adding production notes field

pstjean:


app/representers/api/story_representer.rb
```diff
@@ -5,6 +5,7 @@ class Api::StoryRepresenter < Api::BaseRepresenter
   property :title
   property :clean_title
   property :short_description
+  property :production_notes
```
cavis: Do we want this field to be public?  Could put this in the `auth/story_representer` instead, so it's only accessible when logged in + member-of-account.

pstjean: We definitely don't, thanks for catching. Fixed.

Issue #505, opened 2019-03-05
filter episodes/stories by published state and date range/month


sandikbarr:
The production calendar dashboard view requires that we can filter episodes/stories by published state. Only a single state can be selected at a time, so draft, scheduled, or published. The calendar view additionally requires that we can filter episodes by month and start the list/calendar on a certain date.

PR #513, opened 2019-03-22
Filter by published state

cavis:
For #505.  Adds a new `state` filter to authorized story endpoints.  Filters include (url-encoding the `=`):

- [x] `?filters=state%3Dpublished` (published_at in the past)
- [x] `?filters=state%3Dscheduled` (published_at in the future)
- [x] `?filters=state%3Ddraft` (published_at is null)

Endpoints include:

- [x] `/api/v1/authorization/stories`
- [x] `/api/v1/authorization/accounts/:account_id/stories`
- [x] `/api/v1/authorization/networks/:network_id/stories`
- [x] `/api/v1/authorization/series/:series_id/stories`

And also added to the search endpoints:

- [x] `/api/v1/authorization/stories/search`
- [x] `/api/v1/authorization/series/:series_id/stories/search`

### DIFF
--- a/app/controllers/api/auth/stories_controller.rb
+++ b/app/controllers/api/auth/stories_controller.rb
@@ -8,7 +8,7 @@ class Api::Auth::StoriesController < Api::StoriesController
 
   filter_resources_by :account_id, :series_id, :network_id
 
-  filter_params :highlighted, :purchased, :v4, :text, :noseries
+  filter_params :highlighted, :purchased, :v4, :text, :noseries, :state
 
   sort_params default: { updated_at: :desc },
               allowed: [:id, :created_at, :updated_at, :published_at, :title,
@@ -42,7 +42,20 @@ class Api::Auth::StoriesController < Api::StoriesController
 
   def filtered(resources)
     resources = resources.unseries if filters.noseries?
+    resources = filter_by_state(resources, filters.state) if filters.state?
     super(resources)
+  end
+
+  def filter_by_state(resources, state)
+    if state == 'published'
+      resources.published
+    elsif state == 'scheduled'
+      resources.scheduled
+    elsif state == 'draft'
+      resources.draft
+    else
+      resources.none
+    end
   end
 
   def resources_base
--- a/app/controllers/api/auth/stories_controller.rb
+++ b/app/controllers/api/auth/stories_controller.rb
@@ -20,6 +20,12 @@ class Api::Auth::StoriesController < Api::StoriesController
 
   before_filter :check_user_network, only: [:index], if: -> { params[:network_id] }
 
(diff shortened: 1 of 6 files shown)


app/controllers/api/auth/stories_controller.rb
```diff
@@ -8,7 +8,7 @@ class Api::Auth::StoriesController < Api::StoriesController
 
   filter_resources_by :account_id, :series_id, :network_id
 
-  filter_params :highlighted, :purchased, :v4, :text, :noseries
+  filter_params :highlighted, :purchased, :v4, :text, :noseries, :state
```
cavis: I opted to call this `state` and give it a single value.  The alternative would be completely separate filters like:

```
?filters=published
?filters=scheduled
?filters=draft
```

Open to suggestions on which is better.

docker-compose-ci.yml
```diff
@@ -43,7 +43,7 @@ services:
     volumes:
       - ./test/db/:/docker-entrypoint-initdb.d/
   elasticsearch:
-    image: docker.elastic.co/elasticsearch/elasticsearch-oss:6.3.0
+    image: elasticsearch:6.4.3
```
cavis: Was having issues with that `elastic.co` repo.  So switched to the official dockerhub instance, which is supposedly the same, but not hosted by Elastic.  Also upgraded to 6.4, which is what prod-cms uses right now.

app/controllers/api/auth/stories_controller.rb
```diff
@@ -68,6 +81,19 @@ def search
   def search_params
     sparams = super
     sparams[:fq]['series_id'] = 'NULL' if filters.noseries?
+    sparams[:fq]['published_at'] = search_by_state(filters.state) if filters.state?
     sparams
   end
+
+  def search_by_state(state)
+    if state == 'published'
+      '[* TO now]'
```
cavis: The inclusive/exclusive range syntax is documented [here](https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-query-string-query.html#_ranges).

app/models/story.rb
```diff
@@ -122,7 +122,8 @@ def self.paranoia_scope
       select('pieces.*, series.subscription_approval_status, series.subscriber_only_at')
   }
   scope :published, -> { where('`published_at` <= ?', Time.now) }
-  scope :unpublished, -> { where('`published_at` IS NULL OR `published_at` > ?', Time.now) }
+  scope :scheduled, -> { where('`published_at` > ?', Time.now) }
+  scope :draft, -> { where('`published_at` IS NULL', Time.now) }
```
svevang: unneeded func param?

cavis: ah, good call.

PR #514, opened 2019-03-25
Published/released range filter

cavis:
For #505.  Adds new `?filter`s to the stories/search endpoints, which restricts the range of the `published_at || released_at` timestamps:

- [x] `?filters=before%3D2019-03-01T:00:00:00Z` (published or released set before 3/1, exclusive)
- [x] `?filters=after%3D2019-03-01T:00:00:00Z` (published or released set after 3/1, inclusive)

Note these are added to the _public_ stories endpoints as well.  And the filters can be combined.

Also note that you can just give a date as the start end - which should work for the production calendar view in Publish:

```
/api/v1/authorized/stories?filters=after%3D2019-02-01,before%3D2019-03-01
```

### DIFF
--- a/app/controllers/api/stories_controller.rb
+++ b/app/controllers/api/stories_controller.rb
@@ -7,7 +7,7 @@ class Api::StoriesController < Api::BaseController
 
   filter_resources_by :series_id, :account_id, :network_id
 
-  filter_params :highlighted, :purchased, :v4, :text
+  filter_params :highlighted, :purchased, :v4, :text, before: :time, after: :time
 
   sort_params default: { published_at: :desc, updated_at: :desc },
               allowed: [:id, :created_at, :updated_at, :published_at, :title,
@@ -27,6 +27,7 @@ class Api::StoriesController < Api::BaseController
     sparams[:fq]['network_id'] = params[:network_id] if params[:network_id]
     sparams[:fq]['series_id'] = params[:series_id] if params[:series_id]
     sparams[:fq]['app_version'] = 'v4' if filters.v4?
+    sparams[:fq]['published_released_at'] = search_before_after() if filters.before? || filters.after?
     sparams
   end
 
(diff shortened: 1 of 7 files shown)


test/models/story_test.rb
```diff
@@ -288,6 +288,30 @@
...
+      create(:unpublished_story, title: title, short_description: '1', released_at: now)
+      create(:story, title: title, short_description: '2', published_at: now)
+      create(:story, title: title, short_description: '3', published_at: now + 1.second)
+
+      stories = Story.where(title: title).published_released_before(now + 1.second)
+      stories.pluck(:short_description).must_equal ['1', '2']
+    end
+
+    it 'filters stories published or released after a date' do
+      now = DateTime.parse('2019-03-22T00:00:10Z')
```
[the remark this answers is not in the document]
cavis: Yeah, but by `DateTime.parse('2019-03-22')` comes out in UTC by default, where `Time.parse` does not.

app/controllers/api/stories_controller.rb
```diff
@@ -7,7 +7,7 @@ class Api::StoriesController < Api::BaseController
 
   filter_resources_by :series_id, :account_id, :network_id
 
-  filter_params :highlighted, :purchased, :v4, :text
+  filter_params :highlighted, :purchased, :v4, :text, before: :time, after: :time
```
cavis: I used `before` and `after` as the filter names - which is a bit purposefully-ambiguous.  Thinking that at some point we also might make this search `updated_at` for stories without a published/released date set.

But generally - I'm more okay with these simplistic names for a `?filter=<name>` rather than as a query param of their own.

app/controllers/api/stories_controller.rb
```diff
@@ -112,13 +113,25 @@ def resources_base
...
     if highlighted?
       resources
     else
       super
     end
   end
 
+  def search_before_after
+    if filters.before? && filters.after?
+      "[#{filters.after.iso8601} TO #{filters.before.iso8601}}"
```
cavis: These -> [ES range queries](https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-query-string-query.html#_ranges)

app/controllers/concerns/api_filtering.rb
```diff
@@ -8,11 +8,17 @@ module ApiFiltering
...
       class_attribute :allowed_filter_names
+      class_attribute :allowed_filter_types
     end
   end
 
   class UnknownFilterError < NoMethodError
   end
+  class BadFilterValueError < HalApi::Errors::ApiError
+    def initialize(msg)
+      super(msg, 400)
```
cavis: I wanted these to be an HTTP 400, which is already built into the HalApi error handling.

app/controllers/concerns/api_filtering.rb
```diff
@@ -74,4 +102,16 @@ def parse_filters_param
...
   end
+
+  def parse_date(str)
+    Date.parse(str)
+  rescue ArgumentError
+    raise BadFilterValueError.new "Invalid filter date: '#{str}'"
+  end
+
+  def parse_time(str)
+    Time.find_zone('UTC').parse(str) || (raise ArgumentError.new 'Nil result!')
```
cavis: Oddly ... `Time.parse('bad')` throws an error, but `Time.find_zone('UTC').parse('bad')` returns nil.

PR #523, opened 2019-06-17
Update audio file account

cavis:


app/models/audio_file.rb
```diff
@@ -32,7 +32,7 @@ class AudioFile < BaseModel
       self.status ||= UPLOADED
     end
 
-    if !account && story
+    if !account_id && story.try(:account_id)
```
cavis: Was seeing a testing bug where the `account_id` would change, but `story.account` was nil (and refused to auto-load).  Odd test failures!

app/models/series.rb
```diff
@@ -196,6 +197,11 @@ def set_app_version
   end
 
   def update_account_for_stories
-    stories.each { |s| s.update_attributes!(account_id: account_id) } if account_id_changed?
+    if account_id_changed?
+      stories.update_all(account_id: account_id)
+      audio_files.with_deleted.unscope(where: :promos).
+        reorder('').update_all(account_id: account_id)
```
cavis: These should run faster, and avoid unnecessary callbacks (it kicks off search indexing itself at the end of this).

Basically, these update_alls end up running these 2 queries:

```
UPDATE `pieces` 
SET `pieces`.`account_id` = 123 
WHERE (`pieces`.`deleted_at` IS NULL OR `pieces`.`app_version` = 'v4') AND `pieces`.`series_id` = 183 
ORDER BY episode_number DESC, position DESC, published_at DESC
```

```
UPDATE `audio_files` 
  INNER JOIN `audio_versions` ON `audio_files`.`audio_version_id` = `audio_versions`.`id` 
  INNER JOIN `pieces` ON `audio_versions`.`piece_id` = `pieces`.`id` 
SET `audio_files`.`account_id` = 123 
WHERE (`pieces`.`deleted_at` IS NULL OR `pieces`.`app_version` = 'v4') AND `pieces`.`series_id` = 183
```

(Removing the order-by clause prevents a sql "ambiguous column name 'position'" error).

cavis: Can't hurt - good call!

PR #524, opened 2019-06-28
Feat/ics export

svevang:
Targeting https://github.com/PRX/publish.prx.org/issues/586 this PR sets up a production calendar export for a series in the ical format.

### DIFF
--- a/app/models/story.rb
+++ b/app/models/story.rb
@@ -163,7 +163,11 @@ class Story < BaseModel
     network_visible.
       series_visible.
       coalesce_published_released('ASC').
-      having('pieces_published_released_at is not null')
+      order('pieces_published_released_at ASC').
+      having('pieces_published_released_at is not null').
+      pluck("COALESCE(published_at, released_at) AS pieces_published_released_at, \
+              season_identifier, \
+              episode_identifier")
   end
 
   def self.text_search(text, params = {}, authz = nil)
--- a/app/models/story.rb
+++ b/app/models/story.rb
@@ -159,6 +159,13 @@ class Story < BaseModel
 
   scope :public_stories, -> { published.network_visible.series_visible }
 
+  scope :public_calendar_stories, -> do
+    network_visible.
+      series_visible.
+      coalesce_published_released('ASC').
+      having('pieces_published_released_at is not null')
+  end
+
   def self.text_search(text, params = {}, authz = nil)
     builder = StoryQueryBuilder.new(query: text, params: params, authorization: authz)
     search(builder.as_dsl).records
--- a/app/representers/api/series_representer.rb
+++ b/app/representers/api/series_representer.rb
@@ -9,13 +9,8 @@ class Api::SeriesRepresenter < Api::BaseRepresenter
   property :created_at, writeable: false
   property :updated_at, writeable: false
   property :app_version, writeable: false
-  property :calendar_ics_url, writeable: false, exec_context: :decorator
 
   alternate_link
-
-  def calendar_ics_url
-    api_series_calendar_url(represented, format: :ics)
-  end
 
   link :stories do
     {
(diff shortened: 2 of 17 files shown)


docker-compose.yml
```diff
@@ -23,6 +23,9 @@ services:
     environment:
       VIRTUAL_HOST: cms.prx.docker
       LOCAL_ENV: "true"
+      FEEDER_DB_HOST: pg
```
cavis: Are these really required here?  I thought they were only used by the feeder import script.  But maybe the tests need them or something.

app/representers/api/series_representer.rb
```diff
@@ -9,9 +9,14 @@ class Api::SeriesRepresenter < Api::BaseRepresenter
   property :created_at, writeable: false
   property :updated_at, writeable: false
   property :app_version, writeable: false
+  property :calendar_ics_url, writeable: false, exec_context: :decorator
```
cavis: Shouldn't this be a `_link`, not a property?

app/models/story.rb
```diff
@@ -159,6 +159,13 @@ def self.paranoia_scope
 
   scope :public_stories, -> { published.network_visible.series_visible }
 
+  scope :public_calendar_stories, -> do
```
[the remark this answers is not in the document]
cavis: Hmm... I'm wondering if that's a security issue.  Would there be any problem with just throwing the calendar under the `/authorization/series` endpoints?

docker-compose.yml
svevang: Yep,  the tests in docker wont run unless these are set.

app/models/story.rb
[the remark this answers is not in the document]
cavis: Update: does need to be a publicly accessible URL.  But should leave the story.title out, for security reasons.  Instead, use the Season/Episode number, falling back to the drop date.

app/representers/api/series_representer.rb
svevang: The gist is that you can't "follow" this link (e.g. using the PRXAccess gem), you can't inspect its attributes etc.. It's not a HAL doc. So it seems to more closely follow as an attribute of the model.

docker-compose.yml
cavis: +1

app/representers/api/series_representer.rb
cavis: Hmm... that feels like a shortcoming of the gem.  You should be able to get the URL for a link without following it.  How do we currently handle non-json `type`d links, like images?

```json
{
  "prx:image": {
    "id": 8457,
    "filename": "hermoney-3000x3000-768x768.jpg",
    "size": 109018,
    "_links": {
      "self": {
        "href": "/api/v1/accounts/217417/image",
        "profile": "http://meta.prx.org/model/image/account"
      },
      "profile": {
        "href": "http://meta.prx.org/model/image/account"
      },
      "enclosure": {
        "href": "https://cms.prx.org/pub/1c4c1a73d04f60aaac090731ac1c0c65/0/web/account_image/8457/medium/hermoney-3000x3000-768x768.jpg",
        "type": "image/jpeg"
      },
      "original": {
        "href": "https://cms.prx.org/pub/b56e9d6e928725337d987fa5c3d673f1/0/web/account_image/8457/original/hermoney-3000x3000-768x768.jpg",
        "type": "image/jpeg"
      }
    }
  }
}
```

cavis: Or maybe I misunderstood your comment.  But in any case, having a non-HAL mime in the link `type` is valid per https://tools.ietf.org/html/draft-kelly-json-hal-06#section-5.3

app/services/public_calendar_title_placeholder_builder.rb
```diff
@@ -0,0 +1,34 @@
+class PublicCalendarTitlePlaceholderBuilder
```
cavis: Service, not a concern?  (I don't actually have a strong opinion)

Issue #526, opened 2019-08-15
story publish state filter to support multiple values


sandikbarr:
The Publish home page needs to display a server side paged list of episodes that are in either "draft" or "scheduled" state. To filter those episodes, we use the publish state filter, but it currently only supports one value at a a time. Enhance to support multiple values in the story state filter.

sandikbarr:
I also need a way to include episodes without a release date set in the list for Scheduled and Draft episodes after the current date, so using both the `after` and the `state` params. These undated episodes should always appear at the top of the list when sorted by date ascending.

PR #527, opened 2019-08-21
Unpublished filter/sorting

cavis:
For #526

- [x] Add an `afternull` filter, similar to `after` but also returns episodes with a null published_released_at
- [x] Allow sorting auth-stories by `released_at`, but always returning nulls first (similar to the auth-stories `published_at` sort).
- [x] Add a `state=unpublished` filter value, which allows _either_ `published_at > now()` or `published_at = null`
    _NOTE: this required some changes to how the **?state** filter was applied to elasticsearch.  Instead of using/abusing the `FieldedSearchQuery` class, this sets up ES filters directly through the `StoryQueryBuilder`._

### DIFF
--- a/app/controllers/api/auth/stories_controller.rb
+++ b/app/controllers/api/auth/stories_controller.rb
@@ -12,8 +12,8 @@ class Api::Auth::StoriesController < Api::StoriesController
                 before: :time, after: :time, afternull: :time
 
   sort_params default: { updated_at: :desc },
-              allowed: [:id, :created_at, :updated_at, :published_at, :title,
-                        :episode_number, :position, :published_released_at]
+              allowed: [:id, :created_at, :updated_at, :published_at, :released_at,
+                        :title, :episode_number, :position, :published_released_at]
 
   announce_actions :create, :update, :destroy, :publish, :unpublish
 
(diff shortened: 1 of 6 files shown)


app/controllers/api/auth/stories_controller.rb
```diff
@@ -82,17 +86,13 @@ def search
   def search_params
     sparams = super
     sparams[:fq]['series_id'] = 'NULL' if filters.noseries?
-    sparams[:fq]['published_at'] = search_by_state(filters.state) if filters.state?
+    sparams[:state] = search_by_state(filters.state) if filters.state?
```
cavis: This is the change to how the ES json is built ... instead of abusing the `:fq` fielded search, this just manually sets up the range/bool filters.

app/services/story_query_builder.rb
```diff
@@ -50,6 +57,41 @@ def published_filter
     end
   end
 
+  def state_filter
+    if params[:state] === 'published'
+      Filter.new do
+        range published_at: { lte: 'now' }
+      end
+    elsif params[:state] === 'unpublished'
+      Filter.new do
```
cavis: And here is where the actual ES filters are built.  Range filters are pretty simple ... but for this "unpublished" one I needed to combine (with a `bool`) the must-not-exist OR range.

PR #550, opened 2020-03-12
Fix template segment counts

cavis:
The only place `audio_version_templates.segment_count` was being set was the feed importer.  So we weren't running seg-count validations.

- [x] Reset `segment_count` every time a child `AudioFileTemplate` changes
  - We were already touching the timestamp field, so this change keeps doing that but _also_ checks the AudioVersionTemplate.audio_file_templates.count
  - You can still manually use/set the `segment_count` field, _IFF_ you have no child file_templates
- [ ] Add the `test_after_commit` gem so these after_commit hooks actually run in the tests.  (`database_cleaner` usually prevents them).
- [x] Add the `m` gem because it's convenient
- [x] Send the story `released_at` date in the initial feeder POST episode
- [x] Add `segment_count` to audio_version and audio_version_template representers

### DIFF
--- a/app/models/story_distributions/episode_distribution.rb
+++ b/app/models/story_distributions/episode_distribution.rb
@@ -94,6 +94,7 @@ class StoryDistributions::EpisodeDistribution < StoryDistribution
       content: story.description_html,
       categories: story.tags,
       published_at: story.published_at,
+      released_at: story.released_at,
       updated_at: story.updated_at,
       url: self.class.default_story_url(story)
     }
--- a/test/models/audio_version_template_test.rb
+++ b/test/models/audio_version_template_test.rb
@@ -78,10 +78,17 @@ describe AudioVersionTemplate do
     audio_version_template.audio_file_templates.count.must_equal 0
     audio_version_template.segment_count.must_be_nil
 
-    audio_version_template.audio_file_templates.create!(label: 'seg1')
-    audio_version_template.audio_file_templates.create!(label: 'seg2')
+    f1 = audio_version_template.audio_file_templates.create!(label: 'seg1')
+    f1.run_callbacks(:commit)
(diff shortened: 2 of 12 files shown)


app/models/story_distributions/episode_distribution.rb
```diff
@@ -94,6 +94,7 @@ def episode_attributes
       content: story.description_html,
       categories: story.tags,
       published_at: story.published_at,
+      released_at: story.released_at,
```
cavis: Note to self: Feeder needs to know about this new POST attribute.

cavis: Updated note to self: apparently feeder doesn't care if you POST an extra attribute.  It just ignores it.

kookster: Yes, that's the design, it is not using some schema to validate, so fine to update CMS first, then feeder.

test/models/audio_version_template_test.rb
```diff
@@ -74,4 +74,21 @@
     audio_version_template.validate_audio_file(audio_file).must_include 'is not in video format'
   end
 
+  it 'updates segment count when child file templates change' do
+    audio_version_template.audio_file_templates.count.must_equal 0
+    audio_version_template.segment_count.must_be_nil
+
+    f1 = audio_version_template.audio_file_templates.create!(label: 'seg1')
+    f1.run_callbacks(:commit)
```
kookster: I seem to recall this is how we handle these in tests, though I applaud your efforts to get rid of needing to invoke.

app/models/story_distributions/episode_distribution.rb
kookster: ⚠️ Change request - I believe you also need to add `released_at` to the story representer, so async messages and api responses include it as well.

PR #551, opened 2020-03-26
Process Audio Files with Porter, use Single Porter Job pattern for Audio + Images

chrisrhoden:
This PR is mostly a vehicle to review general direction and make sure I am not going off the deep end. I will make a new one when the other branch gets merged


app/models/audio_file.rb
```diff
@@ -94,4 +99,21 @@ def set_status
       self.status_message = errors
     end
   end
+
+  def analyze!
+    return if status != UPLOADED
+
+    submit_porter_job "cms::audio_file:analyze:#{id}", asset_url do
```
cavis: Ooh, interesting.  This reminds me of rails Global Ids ... any reason not to use those?  I guess they don't include any notion of which job they were for (`analyze`), unless we extended them in some way.  But at the moment, it's only 1 job-type per model right?

```ruby
> UserImage.last.to_global_id.to_s
=> "gid://prx/UserImage/43920"
> GlobalID::Locator.locate 'gid://prx/UserImage/43920'
=> #<UserImage id: 43920, ...>
```

And crazy idea: move images over to GIDs and we can drop that porter_job_id field?  (It's easier to get rid of before this makes it to prod).

cavis: Though I do wonder if this could cause problems with order-of-callbacks.

app/models/concerns/portered.rb
```diff
@@ -0,0 +1,79 @@
...
+    def format_callback(callback)
+      if callback[:sqs].present?
+        {
+          Type: 'AWS/SQS',
+          Queue: URI::HTTPS.build(
+            host: "sqs.#{ENV['AWS_REGION']}.amazonaws.com",
+            path: "/#{ENV['AWS_ACCOUNT_ID']}/#{callback[:sqs]}"
+          ).to_s
+        }
+      elsif callback[:s3].present?
```
cavis: Nice - you're using this for testing, i presume?

app/models/audio_file.rb
```diff
@@ -94,4 +99,21 @@ def set_status
       self.status_message = errors
     end
   end
+
+  def analyze!
```
cavis: Would it make more sense to call this `process!` ... since it's doing both a copy and an analyze?

app/workers/audio_callback_worker.rb
```diff
@@ -63,15 +74,84 @@ def update_audio(audio, job)
...
-    Shoryuken.logger.info("Updating #{job['type']}[#{audio.id}]: status => #{end_state}")
-    audio.save!
+    finish(audio)
   end
+  # rubocop:enable Metrics/AbcSize, Metrics/PerceivedComplexity
 
   private
 
+  def finish(audio)
+    # save and announce the audio changes on its story
```
cavis: Wait, where's the "announce"?  It seems to be up in `perform()` ... not down here.  (The old comment also looks inaccurate).

app/workers/image_callback_worker.rb
```diff
@@ -33,20 +39,21 @@ def announce_image_changed(image)
...
-      when 'resize'
-        callback_resize(image, job['JobResult'])
-      else
-        return nil
+      image.with_lock do
+        begin
+          callback_copy(image, job_result)
+          callback_analyze(image, job_result)
+          callback_thumb(image, job_result)
+        rescue ImageProcessingError => error
```
cavis: Ooh, i like this.  Readable++.

app/models/audio_file.rb
cavis: Update: I don't think out-of-order callbacks will be an issue in real life.  These `process!` methods are only called by the controllers, and if we need to debounce a bit could just add a `def process!(force = false)` flag, that normally skips processing if status == UPLOADED.  (Similar to how feeder does this).

app/models/concerns/portered.rb
chrisrhoden: yep!

app/workers/audio_callback_worker.rb
chrisrhoden: ah yes, this is from before when we were not announcing after every callback, will fix.

app/models/image.rb
```diff
@@ -58,106 +48,45 @@ def self.policy_class
...
+
+    tasks = ImageUploader.version_formats.map do |(name, dimensions)|
       derivative = file.public_send(name).path
       {
         Type: 'Image',
-        Format: format,
         Metadata: 'PRESERVE',
         Resize: {
-          Fit: name == 'square' ? 'cover' : 'inside',
+          Fit: 'contain',
```
chrisrhoden: Looks like a bad merge here

chrisrhoden:
@cavis I think that's all of your feedback handled

PR #553, opened 2020-03-27
Chore/image headers in s3

chrisrhoden:


app/models/image.rb
```diff
@@ -83,6 +83,10 @@ def copy_upload!
               Mode: 'AWS/S3',
               BucketName: ENV['AWS_BUCKET'],
               ObjectKey: "#{fixerable_final_path}/#{filename}"
+              ContentType: 'REPLACE',
```
farski: Missing a comma before

PR #558, opened 2020-04-15
Upgrade PrxAuth::Rails gem

### LINKED ISSUE #557: Upgrade prx_auth-rails gem & update code for checking permissions for new API

chrisrhoden:
closes #557 

Not sure if it's just a weird thing with my system or just trying other test suites and getting used to them but the test suite is really dragging for me which cause this to take a long time.

### DIFF
--- a/app/models/authorization.rb
+++ b/app/models/authorization.rb
@@ -20,7 +20,7 @@ class Authorization
   end
 
   def token_auth_accounts
-    token_ids = token.authorized_resources.try(:keys)
+    token_ids = (token.resources(:member) + token.resources(:admin)).uniq
     @token_auth_accounts = Account.where(id: token_ids) if token_ids
   end
 
@@ -41,9 +41,8 @@ class Authorization
   end
 
   def cache_key
-    token_keys = (token.attributes || {}).keys.sort.map { |k| "#{k}:#{token.attributes[k]}" }
     key_components = ['c', self.class.model_name.cache_key]
-    key_components << OpenSSL::Digest::MD5.hexdigest(token_keys.join(','))
+    key_components << OpenSSL::Digest::MD5.hexdigest(token.resources.join(' '))
     ActiveSupport::Cache.expand_cache_key(key_components)
   end
 
--- a/config/initializers/prx_auth.rb
+++ b/config/initializers/prx_auth.rb
@@ -1,0 +1,4 @@
+PrxAuth::Rails.configure do |config|
+  config.install_middleware = ENV['ID_HOST'].blank?
+  config.namespace = :cms
+end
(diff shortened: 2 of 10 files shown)


app/models/authorization.rb
```diff
@@ -20,7 +20,7 @@ def default_account
   end
 
   def token_auth_accounts
-    token_ids = token.authorized_resources.try(:keys)
+    token_ids = (token.resources(:member) + token.resources(:admin)).uniq
```
cavis, line 23: Won't this break as soon as PRX/id.prx.org#166 lands?  I thought the plan was to have our apps just pull account_ids out of the token and disregarding scopes (pretty much what we're doing now anyways) ... until we get some real scopes defined in PRX/internal#523 ?

And then sometime after deploying granular-scoped tokens ... we'd come back and lock down our server apps to inspect the actual scope strings attached to each account.  At which point this would become `token.resources(:read_private)`, and probably rename the method to something more specific like `token_auth_readable_accounts`.

app/models/authorization.rb
```diff
@@ -41,9 +41,8 @@ def authorized?(account)
   end
 
   def cache_key
-    token_keys = (token.attributes || {}).keys.sort.map { |k| "#{k}:#{token.attributes[k]}" }
     key_components = ['c', self.class.model_name.cache_key]
-    key_components << OpenSSL::Digest::MD5.hexdigest(token_keys.join(','))
+    key_components << OpenSSL::Digest::MD5.hexdigest(token.resources.join(' '))
```
cavis, line 45: Oh right ... CMS attempts to cache authorized resources.  Hmmm... I suppose this should still work.

config/initializers/prx_auth.rb
```diff
@@ -0,0 +1,4 @@
+PrxAuth::Rails.configure do |config|
+  config.install_middleware = ENV['ID_HOST'].blank?
```
cavis, line 2: Wait, what?  You install middleware only if the ID_HOST wasn't set?

app/models/authorization.rb
chrisrhoden: I think that for the purposes of compatibility the scope list will continue to include the role-name (admin or member) so that we also don't open the flood gates to bad behavior.

and yes, after deployment of id changes, this has to be updated.

config/initializers/prx_auth.rb
chrisrhoden: yep! the install_middleware flag only needs to be disabled if we are installing our own manually (https://github.com/PRX/cms.prx.org/pull/558/files#diff-b1fe55db50c712fef0673345e5b9c0d9R61 here) - but I also have a ticket in the PrxAuth::Rails gem to allow configuring those options directly through the rails config object so will circle back to that later. https://github.com/PRX/prx_auth-rails/issues/5

PR #566, opened 2020-08-01
Convert permission checking to use token scopes

chrisrhoden:
Modify Policy classes to check for narrow scopes instead of (a) general membership or (b) role-level abilities. Tokens are distinct from profiles and may not include all of the permissions implied by a given role.

### DIFF
--- a/app/controllers/api/series_controller.rb
+++ b/app/controllers/api/series_controller.rb
@@ -35,14 +35,8 @@ class Api::SeriesController < Api::BaseController
 
   def create_resource
     super.tap do |series|
-      if account && authorization.authorized?(account)
-        series.account_id ||= account.id
-      end
-
-      if authorization.authorized?(current_user.default_account)
-        series.account_id ||= current_user.account_id
-      end
-
+      series.account_id ||= account.id if account
+      series.account_id ||= current_user.account_id
       series.account_id ||= authorization.token_auth_accounts.first.try(:id)
     end
   end
--- a/app/controllers/api/stories_controller.rb
+++ b/app/controllers/api/stories_controller.rb
@@ -84,14 +84,10 @@ class Api::StoriesController < Api::BaseController
   end
 
   def create_resource
-    super.tap do |story|
+    @create_resource ||= super.tap do |story|
       story.creator_id = current_user.id
-      story.account_id ||= story.series.try(:account_id)
       story.account_id ||= account.id if account
       story.account_id ||= current_user.account_id
-      story.account_id ||= (
-        authorization.resources(:story) + authorization.resources(:story_draft)
-      ).first
     end
   end
 
--- a/app/controllers/api/stories_controller.rb
+++ b/app/controllers/api/stories_controller.rb
@@ -87,13 +87,9 @@ class Api::StoriesController < Api::BaseController
     super.tap do |story|
       story.creator_id = current_user.id
       story.account_id ||= story.series.try(:account_id)
-
-      if account
-        story.account_id ||= account.id
-      end
-      
+      story.account_id ||= account.id if account
       story.account_id ||= current_user.account_id
-      story.account_id ||= authorization.token_auth_accounts.first.try(:id)
+      story.account_id ||= (authorization.resources(:story) + authorization.resources(:story_draft)).first
     end
   end
 
--- a/app/controllers/api/stories_controller.rb
+++ b/app/controllers/api/stories_controller.rb
@@ -88,14 +88,11 @@ class Api::StoriesController < Api::BaseController
       story.creator_id = current_user.id
       story.account_id ||= story.series.try(:account_id)
 
-      if account && authorization.authorized?(account)
+      if account
         story.account_id ||= account.id
       end
-
-      if authorization.authorized?(current_user.default_account)
-        story.account_id ||= current_user.account_id
-      end
-
+      
+      story.account_id ||= current_user.account_id
       story.account_id ||= authorization.token_auth_accounts.first.try(:id)
     end
   end
--- a/app/controllers/api/stories_controller.rb
+++ b/app/controllers/api/stories_controller.rb
@@ -89,7 +89,9 @@ class Api::StoriesController < Api::BaseController
       story.account_id ||= story.series.try(:account_id)
       story.account_id ||= account.id if account
       story.account_id ||= current_user.account_id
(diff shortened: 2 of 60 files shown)


app/policies/network_membership_policy.rb
```diff
@@ -5,14 +5,12 @@ def create?
...
   end
 
   # destroy by network owner account admin, or by network member account admin
   def destroy?
-    token && (
-      token.authorized?(resource.network.account.id, :admin) ||
-      token.authorized?(resource.account.id, :admin)
-    )
+    token&.authorized?(resource.network.account.id, :network_members) ||
+      token&.authorized?(resource.account.id, :network_members)
```
chrisrhoden, line 14: This feels weird to me - I think network-members is being a little overloaded but I don't have a better suggestion in the current list of scopes - @cavis thoughts?

app/policies/purchase_policy.rb
```diff
@@ -1,6 +1,6 @@
 class PurchasePolicy < ApplicationPolicy
   def create?
-    token && token.authorized?(resource.purchaser_account.id, :admin)
+    token&.authorized?(resource.purchaser_account.id, :purchase)
```
cavis, line 3: Ooh, `purchase`!  That's new!

app/policies/user_policy.rb
```diff
@@ -4,6 +4,6 @@ def create?
   end
 
   def update?
-    token.user_id == resource.id
+    false
```
cavis, line 7: Yeah, `users` may move out of this DB anyways.  So makes sense!

test/representers/api/story_representer_test.rb
```diff
@@ -24,6 +24,15 @@
       d_story.title.must_equal 'title'
       d_story.account_id.must_equal 8
     end
+
+    it 'can set the series' do
```
cavis, line 28: What's this test about?  Just making sure you can move between series you have account access to?

app/policies/accountable_policy.rb
```diff
@@ -1,10 +1,15 @@
 class AccountablePolicy < ApplicationPolicy
+  def initialize(token, resource, scope = :account)
```
cavis, line 2: Does this default `:account` scope get used anywhere?

app/controllers/api/series_controller.rb
```diff
@@ -35,14 +35,8 @@ def filtered(resources)
...
     super.tap do |series|
-      if account && authorization.authorized?(account)
-        series.account_id ||= account.id
-      end
-
-      if authorization.authorized?(current_user.default_account)
-        series.account_id ||= current_user.account_id
-      end
-
+      series.account_id ||= account.id if account
```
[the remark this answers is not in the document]
chrisrhoden: Yes, the policy will always make sure that you have write access to the resource (as defined after the transformations have been applied, so there is no risk here)

app/models/authorization.rb
```diff
@@ -19,30 +20,29 @@ def default_account
...
   end
 
+  def account_ids(scope = nil)
+    token.resources(scope).map(&:to_i)
+  end
+
   def token_auth_accounts
-    token_ids = (token.resources(:member) + token.resources(:admin)).uniq
-    @token_auth_accounts = Account.where(id: token_ids) if token_ids
+    @token_auth_accounts ||= Account.where(id: account_ids(:read_private))
```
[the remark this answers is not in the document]
chrisrhoden: It actually always returns an ActiveRecord scope, which could be empty (Backed by a query like `where 1=0`, which happens automatically if you do `where(id: [])`)

app/models/story.rb
```diff
@@ -184,6 +184,16 @@ def touch_series
     series.try(:touch)
   end
 
+  def series=(series_object)
+    super
+    self.account_id = series.account_id if series&.account_id.present?
```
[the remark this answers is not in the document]
chrisrhoden: yes! Per our conversation, I moved it into the model, since it's technically possible to imagine there are other contexts in which a series might be set, and that should _always_ set the account.

app/policies/accountable_policy.rb
chrisrhoden: Yes, for all of the models that just use `AccountablePolicy` as their policy class

app/policies/distributable_policy.rb
```diff
@@ -0,0 +1,15 @@
+class DistributablePolicy < ApplicationPolicy
+  def initialize(token, resource)
+    super
+    sub_policy_class = resource && Pundit::PolicyFinder.new(resource.distributable.class).policy!
```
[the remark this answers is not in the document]
chrisrhoden: I actually copied this from the OwnedPolicy - not sure if it's ideal here, but I think it makes the most sense.

test/representers/api/story_representer_test.rb
chrisrhoden: Yes - I wrote it when I was trying to debug issues with setting the series via the controller - the issue was ultimately not a problem with the representer, and this was already working fine. I can delete the test if you like! goodness knows this suite doesn't need to be bigger.

app/controllers/api/stories_controller.rb
```diff
@@ -84,19 +84,10 @@ def random
...
-      if account && authorization.authorized?(account)
-        story.account_id ||= account.id
-      end
-
-      if authorization.authorized?(current_user.default_account)
-        story.account_id ||= current_user.account_id
-      end
-
-      story.account_id ||= authorization.token_auth_accounts.first.try(:id)
+      story.account_id ||= account.id if account
```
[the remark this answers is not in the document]
chrisrhoden: picked up in the model, below, as you found.

app/policies/account_policy.rb
```diff
@@ -4,10 +4,10 @@ def create?
   end
 
   def update?
-    token && token.authorized?(resource.id)
+    token&.authorized?(resource.id, :account)
   end
 
   def destroy?
-    token && token.authorized?(resource.id, :admin)
+    token&.authorized?(resource.id, :account_members)
```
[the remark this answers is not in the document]
chrisrhoden: yes - as with network_members, it feels a little overloaded but I think it probably makes sense?

test/representers/api/story_representer_test.rb
cavis: Nah, that makes sense.  What's 1 more test to this gigantic test suite?

PR #573, opened 2020-11-05
Add loudness and true peaks columns to schema

farski:
Blocked by https://github.com/PRX/exchange.prx.org/pull/540. **That migration must be run before this is merged**

Adds schema support in CMS for new loudness and true peaks columns in the DB.


kookster (approved):
lgtm - consistent with the exchange migration PR

PR #575, opened 2020-12-10
Authorization endpoint N+1 fixes

cavis:
The `/api/v1/authorization` loads real slow if you have many accounts.

Made a few improvements:
- [x] Joins/includes on account addresses/images/openers
  - I'm actually not sure why the auth+min account representer includes all this stuff, or if anything relies on that.  Maybe would be better to just remove those by-default embeds from the min representers?
- [x] Remove story counts from auth+min account representers
  - It's unlikely anything was really relying on these counts. 
  - Or maybe there's an easy way to eager-loading N+1 counts that i don't know about
- [x] The address representer had a self_url that reloaded the account

### DIFF
--- a/app/controllers/api/series_controller.rb
+++ b/app/controllers/api/series_controller.rb
@@ -37,7 +37,7 @@ class Api::SeriesController < Api::BaseController
     super.tap do |series|
       series.account_id ||= account.id if account
       series.account_id ||= current_user.account_id
-      series.account_id ||= authorization.token_auth_accounts.first.try(:id)
+      series.account_id ||= authorization.token_auth_account_ids.first
     end
   end
 
--- a/app/models/authorization.rb
+++ b/app/models/authorization.rb
@@ -25,7 +25,9 @@ class Authorization
   end
 
   def token_auth_accounts
-    @token_auth_accounts ||= Account.where(id: account_ids(:read_private))
+    @token_auth_accounts ||= Account.where(id: account_ids(:read_private)).
+                             joins(:address, :image, :opener).
+                             includes(:address, :image, :opener)
   end
 
   def token_auth_stories
--- a/app/models/authorization.rb
+++ b/app/models/authorization.rb
@@ -24,9 +24,13 @@ class Authorization
     token.resources(scope).map(&:to_i)
   end
 
+  def token_auth_account_ids
+    account_ids(:read_private)
+  end
+
   def token_auth_accounts
-    @token_auth_accounts ||= Account.where(id: account_ids(:read_private)).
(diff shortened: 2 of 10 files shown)


app/controllers/api/series_controller.rb
```diff
@@ -37,7 +37,7 @@ def create_resource
     super.tap do |series|
       series.account_id ||= account.id if account
       series.account_id ||= current_user.account_id
-      series.account_id ||= authorization.token_auth_accounts.first.try(:id)
+      series.account_id ||= authorization.token_auth_account_ids.first
```
kookster, line 40: nice one, clearly faster

kookster, line 40: nice one, clearly faster

app/representers/api/address_representer.rb
```diff
@@ -11,6 +11,6 @@ class Api::AddressRepresenter < Api::BaseRepresenter
   property :country
 
   def self_url(address)
-    api_account_address_path(address.account) if address.account
+    api_account_address_path(address.account_id) if address.account_id
```
kookster, line 14: nice

app/models/authorization.rb
```diff
@@ -24,8 +24,14 @@ def account_ids(scope = nil)
...
 
+  def token_auth_account_ids
+    account_ids(:read_private)
+  end
+
   def token_auth_accounts
-    @token_auth_accounts ||= Account.where(id: account_ids(:read_private))
+    @token_auth_accounts ||= Account.where(id: token_auth_account_ids).
+                             left_joins(:address, :image, :opener).
+                             includes(:address, :image, :opener)
```
kookster, line 34: yeah, that's a good plan

app/representers/api/auth/account_min_representer.rb
```diff
@@ -5,8 +5,7 @@ class Api::Auth::AccountMinRepresenter < Api::Min::AccountRepresenter
   link :stories do
     {
       href: "#{api_authorization_account_stories_path(represented)}#{index_url_params}",
-      templated: true,
-      count: represented.stories.count
```
kookster, line 9: pretty sure nothing relies on this - it was a bit of a "standard" to offer counts, but not a big deal

kookster (approved):
lgtm

app/representers/api/auth/account_min_representer.rb
cavis: When we someday go through CMS with a fine toothed comb ... would be nice to standardize on what's included in full vs min representers.  Maybe only full get counts and embedded-by-defaults?

PR #596, opened 2021-10-05
Use Redis instead of memcached

### LINKED ISSUE #595: Replace memcached with Redis

farski:
Closes #595 

Untested


cavis:
I'm a bit less motivated to do this, since we do need a memcached around for CMS.  Going to convert to draft for now.

PR #597, opened 2021-10-12
Use explicit queue names when available

farski:


cavis (approved):
👍 But should merge PRX/Infrastructure#554 first.

PR #598, opened 2021-11-09
Custom sqs/sns prefix

cavis:


Gemfile
```diff
@@ -1,7 +1,7 @@
 source 'https://rubygems.org'
 
 # Bundle edge Rails instead: gem 'rails', github: 'rails/rails'
-gem 'rails', '~> 4.2.11'
+gem 'rails', '~> 4.2.11.3'
```
kookster, line 4: it's a pretty minor release, security fixes, which will make gh bots happier

bin/application
```diff
@@ -8,7 +8,7 @@ ApplicationUsage () {
 }
 
 ApplicationUpdate () {
-  bundle exec rake -vt sqs:create announce:configure_broker db:create
+  bundle exec rake -vt announce:verify_config db:create
```
kookster, line 11: yup, and when I tested locally, this was working, showing correct output in local feeder

config/application.rb
```diff
@@ -66,7 +66,7 @@ class Application < Rails::Application
     end
 
     config.active_job.queue_adapter = :shoryuken
-    config.active_job.queue_name_prefix = Rails.env
+    config.active_job.queue_name_prefix = ENV['ANNOUNCE_RESOURCE_PREFIX'] || ENV['RAILS_ENV']
```
kookster, line 69: I couldn't remember if we were setting this explicitly, this looks right, and I like having the fallback to env, esp for local dev

config/shoryuken.yml
```diff
@@ -10,8 +10,8 @@ concurrency:         <%= ENV['WORKER_COUNT'] || 25 %>
...
-  - [<%= env %>_cms_podcast_import, 4]
-  - [<%= env %>_cms_default, 4]
-  - [<%= env %>_cms_audio_callback, 3]
-  - [<%= env %>_cms_image_callback, 2]
-  - [<%= env %>_cms_search_indexer, 1]
+  - [<%= prefix %>_cms_podcast_import, 4]
+  - [<%= prefix %>_cms_default, 4]
+  - [<%= prefix %>_cms_audio_callback, 3]
+  - [<%= prefix %>_cms_image_callback, 2]
+  - [<%= prefix %>_cms_search_indexer, 1]
```
kookster, line 17: lgtm, names all match, priorities the same too

env-example
```diff
@@ -1,3 +1,4 @@
+ANNOUNCE_RESOURCE_PREFIX=development
```
kookster, line 1: we could change this name later if we really want to, but this will work

kookster (approved):
lgtm!

PR #611, opened 2022-06-27
Resource for listing series story tags

cavis:
Bad idea, or baddest idea?


kookster (changes requested):
I think it's a reasonable approach, if you want a separate endpoint instead of including in the series resource itself.

The only thing I'd change is the doing a select distinct on name, or grouping, just to do the deduping work in the query

kookster:
Lgtm

kookster (approved):
Lgtm

PR #621, opened 2023-05-19
Increase public asset expirations

cavis:
Change expiration from 10 minutes to 24 hours.

May help with PRX/beta.prx.org#361.


kookster (approved):
lgtm!

PR #623, opened 2023-08-15
Indicate if user has Feeder UI access

cavis:
Related to PRX/id.prx.org#363.

Return an attribute on the Authorization so indicate if this user has account-access to new Feeder UI.  Publish can then use this to display/redirect the user over to Feeder.

Could honestly be deployed at any time.  Just has to precede Publish changes.

### DIFF
--- a/app/models/authorization.rb
+++ b/app/models/authorization.rb
@@ -16,6 +16,17 @@ class Authorization
     default_account.id
   end
 
+  def feeder_ui_access
+    if (token.resources(:feeder, :read_private) - token_auth_account_ids).any?
+      host = ENV["FEEDER_HOST"]
+      if host.present?
+        host.starts_with?("http") ? host : "https://#{host}"
+      else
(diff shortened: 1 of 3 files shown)


app/models/authorization.rb
```diff
@@ -16,6 +16,17 @@ def id
     default_account.id
   end
 
+  def feeder_ui_access
+    if (token.resources(:feeder, :read_private) - token.resources(:cms, :read_private)).any?
```
farski, line 20: Based on my understanding of what this is doing, `token.resources(:feeder, :read_private).any? && token.resources(:cms, :read_private).empty?` may be more obvious, but I think it does that same thing.

farski (approved):

app/models/authorization.rb
cavis: Ah true, since in `roles.yml` ... every role that has `cms:read-private` also has `feeder:read-private`.

PR #626, opened 2023-09-25
Remove episode url default

radical-ube:
part of [Feeder migration to remove beta links](https://github.com/PRX/feeder.prx.org/issues/825)


cavis (approved):
👍

PR #641, opened 2025-09-02
Authenticate everybody

cavis:
Require creds for all API requests.

Note that the `PublicAssetsController` is outside of this, so will still work without creds.  But that should be fine - can't get to those without some other CMS link sending you there.

### DIFF
--- a/test/controllers/api/account_images_controller_test.rb
+++ b/test/controllers/api/account_images_controller_test.rb
@@ -5,6 +5,8 @@ describe Api::AccountImagesController do
   let(:account_image) { create(:account_image) }
   let(:account) { account_image.account }
 
+  around { |test| @controller.stub(:current_user, true) { test.call } }
+
   it 'should show' do
     get(:show, api_request_opts(account_id: account_image.account_id))
     assert_response :success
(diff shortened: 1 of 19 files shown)


test/controllers/api/accounts_controller_test.rb
```diff
@@ -8,23 +8,27 @@
   let (:write_token) { StubToken.new(nil, nil, user_without_account.id) }
   let (:token) { StubToken.new(nil, ['cms:account'], user.id) }
 
-  it 'should show' do
-    get(:show, { api_version: 'v1', format: 'json', id: account.id })
-    assert_response :success
-  end
+  describe 'with any user' do
+    around { |test| @controller.stub(:current_user, true) { test.call } }
```
cavis, line 12: Some of these tests, I could just add the `around` at the top.

But others test both with/without auth.  So I had to wrap big chunks in a `describe` block.  Diff might be clearer ignoring whitespace on these.

test/controllers/api/account_images_controller_test.rb
```diff
@@ -5,6 +5,8 @@
   let(:account_image) { create(:account_image) }
   let(:account) { account_image.account }
 
+  around { |test| @controller.stub(:current_user, true) { test.call } }
```
kookster, line 8: a good way to fix tests!

kookster (approved):
lgtm - we'll see what it breaks!

